PR #1, opened 2017-02-21
Fb distributions

labkey-susanh:


src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
```diff
@@ -97,11 +100,11 @@ class DistributionExtension
...
 
-    String dir
-    String modulesDir
-    String installerSrcDir
-    String extraSrcDir
-    String archiveDataDir
+    String dir = "${project.rootProject.projectDir}/dist"
+    String modulesDir = "${project.rootProject.buildDir}/distModules"
+    String installerSrcDir = "${project.rootProject.projectDir}/server/installer"
+    String extraSrcDir = "${project.rootProject.buildDir}/distExtra"
```
Kristjansson: How is project instantiated here?

src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
```diff
@@ -82,11 +100,11 @@ class DistributionExtension
     public static final String DIST_FILE_NAME = "distribution"
     public static final String VERSION_FILE_NAME = "VERSION"
 
-    String dir
-    String modulesDir
-    String installerSrcDir
-    String extraSrcDir
-    String archiveDataDir
+    String dir = "${project.rootProject.projectDir}/dist"
```
Kristjansson, line 103: How is project instantiated here?

labkey-susanh: I don't actually know, but it does work...

src/main/groovy/org/labkey/gradle/util/ModuleFinder.groovy
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+
+    @Override
+    FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) {
+        if (dir.getFileName().toString().startsWith(".") ||
+                excluded.contains(dir.getFileName()) ||
+                NON_MODULE_DIRS.contains(dir.getFileName().toString()))
+        {
+            FileVisitResult.SKIP_SUBTREE
```
Kristjansson: Should this have an explicit return statement? If I remember right, groovy will return it anyway because it's the last statement in the function, but adding return would add clarity.

labkey-susanh: Yes, explicit is better.  Done.

Kristjansson: Awesome, thanks!

src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
Kristjansson: Huh, ok sounds good.

Kristjansson (approved):
Looks good!

PR #5, opened 2017-05-12
Spec ID: 29080

Kristjansson:
- Migrates ant create_module to gradle createModule

### DIFF
--- a/src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
+++ b/src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
@@ -8,7 +8,8 @@ import java.util.regex.Matcher
 import java.util.regex.Pattern
 
 /**
- * Created by Joe on 5/11/2017.
+ * Creates a new module based on the contents of trunk/server/moduleTemplate. Sources properties from commandline or
+ * user input based on the prompt flag.
  */
 class CreateModule extends DefaultTask
 {
--- a/src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
+++ b/src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
@@ -1,6 +1,7 @@
 package org.labkey.gradle.task
 
 import org.gradle.api.DefaultTask
+import org.gradle.api.GradleException
 import org.gradle.api.file.CopySpec
 import org.gradle.api.tasks.TaskAction
 
@@ -17,11 +18,11 @@ class CreateModule extends DefaultTask
 
     @TaskAction
     void createModule() {
-        String moduleName = null
-        String moduleDestination = null
-        boolean hasManagedSchema = true
-        boolean createTestFiles = true
-        boolean createApiFiles = true
+        String moduleName = project.hasProperty('moduleName') ? project.moduleName : null
+        String moduleDestination = project.hasProperty('moduleDestination') ? project.moduleDestination : null
+        boolean hasManagedSchema = project.hasProperty('noManagedSchema')
+        boolean createTestFiles = project.hasProperty('createTestFiles')
+        boolean createApiFiles = project.hasProperty('createApiFiles')
 
         if (prompt) {
             project.ant.input(
@@ -37,7 +38,7 @@ class CreateModule extends DefaultTask
             moduleDestination = ant.new_moduleDestination
 
             project.ant.input(
-                    message: "\nWill this module create and manage a database schema? (y/N)",
+                    message: "\nWill this module create and manage a database schema? (Y/n)",
                     addProperty: "new_hasManagedSchema"
             )
             hasManagedSchema = !(ant.new_hasManagedSchema.toLowerCase().equals("n"))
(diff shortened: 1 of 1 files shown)


src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,169 @@
...
+                createApiFiles = project.createApiFiles.toLowerCase().equals("y")
+            }
+        }
+
+        if (moduleName == null) {
+            project.logger.error("moduleName is not specified")
+        }
+        if (moduleDestination == null) {
+            project.logger.error("moduleDestination is not specified")
+        }
```
labkey-susanh: We should probably throw an exception if one of these is null since we can't create a module without these.

src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,169 @@
...
+
+        boolean shouldCreate = true
+
+        File existingModuleDir = new File(moduleDestination)
+        if (existingModuleDir.exists()) {
+            ant.input(
+                    message: "\nModule directory already exists (${moduleDestination}). " +
+                            "Create module anyway? (y/N)\"",
```
labkey-susanh: Maybe we should update this prompt to let folks know the existing one will be deleted if they say "y".  Something like "Delete existing directory and create module anyway?"

src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,169 @@
...
+                    message: "\nEnter the full path for where to put the new module: ",
+                    addProperty: "new_moduleDestination"
+            )
+            moduleDestination = ant.new_moduleDestination
+
+            project.ant.input(
+                    message: "\nWill this module create and manage a database schema? (y/N)",
+                    addProperty: "new_hasManagedSchema"
+            )
+            hasManagedSchema = !(ant.new_hasManagedSchema.toLowerCase().equals("n"))
```
labkey-susanh: The default value here will be "yes", which is consistent with the ant behavior, but the prompt is misleading.  I always interpreted the capitalized one as the default one, so perhaps we should change this to "Will this module create and manage a database schema? (Y/n)".

src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,169 @@
...
+
+    void createNewModule(String moduleName, String moduleDestination, boolean hasManagedSchema, boolean createTestFiles, boolean createApiFiles)
+    {
+        try {
+            project.mkdir(moduleDestination)
+        }
+        catch (Exception e) {
+            project.logger.error("Failed to create new module directory at ${(new File(moduleDestination)).getAbsolutePath()}")
+            project.logger.error("Error generated : ${e.message}")
+            project.logger.error(e.stackTrace.toArrayString())
```
labkey-susanh: This should also throw or return at this point since we don't want to do the substitutions if we don't have a directory.

src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,169 @@
...
+                }
+                return line
+            })
+        })
+        //copy.rename only looks at file names, rather than files and directories.
+        renameCrawler(project.file(moduleDestination), substitutions)
+        File codeOnlyModule = new File("${moduleDestination}/src/org/labkey/${moduleName}/MODULE_NAMECodeOnlyModule.java")
+        if (codeOnlyModule.exists()) {
+            codeOnlyModule.renameTo(new File("${moduleDestination}/src/org/labkey/${moduleName}/MODULE_NAMEModule.java"))
+        }
```
labkey-susanh: Isn't this too late?  Won't the renameCrawler already have renamed these files?  If not, shouldn't it?

labkey-susanh (changes requested):

src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,169 @@
...
+{
+    boolean prompt = true
+
+    @TaskAction
+    void createModule() {
+        String moduleName = null
+        String moduleDestination = null
+        boolean hasManagedSchema = true
+        boolean createTestFiles = true
+        boolean createApiFiles = true
```
labkey-susanh: Set these values from the properties, if they exist. Then check for all the required properties to determine if the prompt is needed instead of passing that in as a parameter.  

The default values for createTestFiles and createApiFiles should be false to be consistent with the ant task.

src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,169 @@
...
+                moduleDestination = project.moduleDestination
+            }
+            if (project.hasProperty('hasManagedSchema')) {
+                hasManagedSchema = !(project.hasManagedSchema.toLowerCase().equals("n"))
+            }
+            if (project.hasProperty('createTestFiles')) {
+                createTestFiles = project.createTestFiles.toLowerCase().equals("y")
+            }
+            if (project.hasProperty('createApiFiles')) {
+                createApiFiles = project.createApiFiles.toLowerCase().equals("y")
```
labkey-susanh: For these boolean properties, let's not require a value.  Instead, just check for the existence of the property.  Then someone can use them on the command line like so:
-PcreateTestFiles -PnoManagedSchema

src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,152 @@
...
+ */
+class CreateModule extends DefaultTask
+{
+    boolean prompt = true
+
+    @TaskAction
+    void createModule() {
+        String moduleName = project.hasProperty('moduleName') ? project.moduleName : null
+        String moduleDestination = project.hasProperty('moduleDestination') ? project.moduleDestination : null
+        boolean hasManagedSchema = project.hasProperty('noManagedSchema')
```
labkey-susanh: This should be negated.

src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,152 @@
...
+        boolean createTestFiles = project.hasProperty('createTestFiles')
+        boolean createApiFiles = project.hasProperty('createApiFiles')
```
labkey-susanh: Instead of using the boolean properties, how about changing to a different kind of property with a value, such as: -Pfiles=test,api,schema.  

Then we extract the boolean values from this property and if there is no such property we know we need to prompt for answers. Then we don't need two tasks.  (I think probably only TeamCity uses the noPrompt task, so it's not a big deal, but if we can make it more friendly to humans to use, they may do so, and then we don't have special logic just for TeamCity that we might forget it needs.)

src/main/groovy/org/labkey/gradle/task/CreateModule.groovy
```diff
@@ -0,0 +1,152 @@
...
+                copy.exclude("**/MODULE_NAMEModule.java")
+            }
+            if (!createTestFiles) {
+                copy.exclude("test/**")
+            }
+            if (!createApiFiles) {
+                copy.exclude("api-src/**")
+            }
+
+            copy.exclude("**/*.iml")
```
labkey-susanh: This can be removed as I believe the .iml files have been removed.

labkey-susanh (changes requested):

labkey-susanh (approved):

PR #6, opened 2017-05-15
Don't flatten pipeline tool directories

labkey-tchad:


src/main/groovy/org/labkey/gradle/task/DeployApp.groovy
```diff
@@ -126,9 +126,20 @@ class DeployApp extends DefaultTask
     // Use this method to preserve file permissions, since ant.copy does not, but this does not preserve last modified times
     private void deployBinariesViaProjectCopy(String osDirectory)
     {
-        project.copy { CopySpec copy ->
-            copy.from project.fileTree("${project.labkey.externalDir}/${osDirectory}").files
-            copy.into "${project.serverDeploy.binDir}"
+        File parentDir = new File("${project.labkey.externalDir}", "${osDirectory}")
```
labkey-susanh, line 129: You don't need the "${..}" unless you are constructing a string that needs these variable values incorporated. You can just do:
 File parentDir = new File(project.labkey.externalDir, osDirectory)

src/main/groovy/org/labkey/gradle/task/DeployApp.groovy
```diff
@@ -126,9 +126,20 @@ class DeployApp extends DefaultTask
...
+            {
+                return pathname.isDirectory()
+            }
+        }
+        for (File dir : subDirs)
+        {
+            project.copy { CopySpec copy ->
+                copy.from dir
+                copy.into "${project.serverDeploy.binDir}"
+            }
```
labkey-susanh, line 142: Doesn't this miss the stuff in the osDirectory itself?

labkey-tchad: Yeah. That's how it worked in ant.

src/main/groovy/org/labkey/gradle/task/DeployApp.groovy
labkey-tchad: Without them it can't decide between the File(String..) and File(File..) constructors.

labkey-susanh: Yay for type ambiguity. :-/

PR #7, opened 2017-09-12
Fb bug fixes

labkey-susanh:


labkey-tchad (approved):
Distribution naming looks good

PR #10, opened 2018-01-17
Fb 1.2.3 snapshot

labkey-susanh:


labkey-tchad (approved):
TeamCity has been running this for a week with no problems. Builds, tests, and installers all look good

PR #14, opened 2018-05-23
Fb 1.2.7 snapshot

labkey-danield:
Updated so ClientApiDistribution has the jdbc jar file.


labkey-danield:
Personal build of the PremiumInstallers completed without any issues on TeamCity.

labkey-susanh (approved):

PR #15, opened 2018-06-05
Fb 1.2.8 snapshot

labkey-danield:
Update to allow for validationOnly TeamCity flag.

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
@@ -229,8 +229,14 @@ class TeamCity extends Tomcat
                         task.dbPropertiesChanged = true
                         task.fn = {
                             properties.mergePropertiesFromFile()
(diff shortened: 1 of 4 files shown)


src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
```diff
@@ -229,8 +229,14 @@ class TeamCity extends Tomcat
                         task.dbPropertiesChanged = true
                         task.fn = {
                             properties.mergePropertiesFromFile()
-                            if (extension.dropDatabase)
-                                SqlUtils.dropDatabase(project, properties)
+                            if (extension.dropDatabase) {
+                                if (extension.getTeamCityProperty("testValidationOnly").equalsIgnoreCase("true")){
```
labkey-susanh: I think you can do this without string comparison using a cast to a Boolean:

(Boolean) getTeamCityProperty("testValidationOnly")

labkey-susanh (approved):

PR #16, opened 2018-06-12
Fb 1.3 snapshot

labkey-susanh:
It might be obvious, but there is no need to review the moduleTemplate files in much depth.  They have been copied from SVN.


labkey-tchad (approved):

labkey-klum (approved):
looks good

PR #18, opened 2018-08-23
Fb 1.3.2 snapshot

labkey-susanh:


moduleTemplate/src/org/labkey/MODULE_DIR_NAME/MODULE_NAMEModule.java
```diff
@@ -40,7 +40,8 @@ public String getName()
     @Override
     public double getVersion()
     {
-        return 17.21;
+        // TODO @@LABKEY_VERSION_NUMBER@@
+        return 18.21;
```
labkey-tchad: Looks like `LABKEY_VERSION_NUMBER` is available. Should this be using it?

labkey-tchad (changes requested):

labkey-tchad (approved):

PR #19, opened 2018-10-17
1.3.3 snapshot

labkey-susanh:


labkey-tchad (approved):
Looks good. Just updated TeamCity trunk project to use this version.

PR #21, opened 2018-10-26
Fb 1.3.4 snapshot

labkey-susanh:


README.md
```diff
@@ -8,6 +8,15 @@ but also make certain assumptions that you may not want to impose on your module
 [LabKey documentation](https://www.labkey.org/Documentation/wiki-page.view?name=gradleModules) for more information.
 
 ## Release Notes
+### version 1.3.4
+*Release*: ???
+(Earliest compatible LabKey version: 18.2)
```
labkey-tchad: Should this be bumped to 18.3 or will we backport today's module.xml fix from trunk?

labkey-tchad: If so, we should remove the test dependency on `com.googlecode.sardine` (TestRunner.groovy:line 104)

labkey-tchad (changes requested):
Looks good. Just one question about the supported labkey version

README.md
labkey-susanh: Yes, should be bumped to 18.3.  I've made these two changes.  I'll let TC run through a build or two before merging.

labkey-tchad (approved):

PR #24, opened 2019-01-14
1.4.0 snapshot

labkey-susanh:
A lot of these changes are to switch to using project.task.register instead of project.task to take advantage of the [task configuration avoidance](https://blog.gradle.org/preview-avoiding-task-configuration-time), which came out of incubation with Gradle 5.1.  

The MultiGit plugin is, as its documentation says, incubating, so has some things left to do still and I likely won't make further changes there for this release unless there are fundamental problems.


src/main/groovy/org/labkey/gradle/plugin/JavaModule.groovy
```diff
@@ -241,14 +258,20 @@ class JavaModule extends FileModule
...
-        // trim nothing from api
-        String apiProjectPath = BuildUtils.getProjectPath(project.gradle, "apiProjectPath", ":server:api")
-        if (project.path.equals(apiProjectPath))
+//        // trim nothing from api
+//        if (BuildUtils.isApi(project))
+//            return config
+//        else // all other modules should remove everything in the dedupe configuration
+//        {
+//            return config - project.configurations.dedupe
+//        }
```
labkey-tchad, line 269: Removed temporarily or for debugging?

src/main/groovy/org/labkey/gradle/plugin/UiTest.groovy
```diff
@@ -77,25 +77,26 @@ class UiTest implements Plugin<Project>
...
         if (project.findProject(schemasProjectPath) != null)
-            BuildUtils.addLabKeyDependency(project: project, config: 'uiTestCompile', depProjectPath: schemasProjectPath, depVersion: project.labkeyVersion)
-        BuildUtils.addLabKeyDependency(project: project, config: 'uiTestCompile', depProjectPath: BuildUtils.getProjectPath(project.gradle, "apiProjectPath", ":server:api"), depVersion: project.labkeyVersion)
-        BuildUtils.addLabKeyDependency(project: project, config: 'uiTestCompile', depProjectPath: BuildUtils.getProjectPath(project.gradle, "remoteApiProjectPath", ":remoteapi:java"), depVersion: project.labkeyVersion)
+            BuildUtils.addLabKeyDependency(project: project, config: 'uiTestImplementation', depProjectPath: schemasProjectPath, depVersion: project.labkeyVersion)
+        BuildUtils.addLabKeyDependency(project: project, config: 'uiTestImplementation', depProjectPath: BuildUtils.getApiProjectPath(project.gradle), depVersion: project.labkeyVersion)
+        BuildUtils.addLabKeyDependency(project: project, config: 'uiTestImplementation', depProjectPath: BuildUtils.getRemoteApiProjectPath(project.gradle), depVersion: project.labkeyVersion)
```
labkey-tchad: Could we remove the schemas and API dependencies from the plugin and define them in SVN instead? We're working on removing test dependencies on product code and it would be nice to get the plugin update portion of the story out of the way now.

src/main/groovy/org/labkey/gradle/plugin/UiTest.groovy
```diff
@@ -77,25 +77,26 @@ class UiTest implements Plugin<Project>
...
-        project.tasks.uiTests.mustRunAfter(project.project(":server").tasks.pickPg)
-        project.tasks.uiTests.mustRunAfter(project.project(":server").tasks.pickMSSQL)
+        project.tasks.register("moduleUiTests", RunUiTest) {
+            RunUiTest task ->
+                task.group = GroupNames.VERIFICATION
+                task.description = "Run UI (Selenium) tests for this module"
+        }
+
+        project.tasks.moduleUiTests.mustRunAfter(project.project(":server").tasks.pickPg)
+        project.tasks.moduleUiTests.mustRunAfter(project.project(":server").tasks.pickMSSQL)
```
labkey-tchad: Should these `mustRunAfter`s be inside the `RunUiTest` closure above? I noticed that pattern earlier.

labkey-tchad (changes requested):

src/main/groovy/org/labkey/gradle/plugin/JavaModule.groovy
labkey-susanh: Commented out because it doesn't work as expected.  I posted to the Gradle forums about this but haven't heard anything.  I need to see if I can tease out a simple repro case so I can file a bug with the Gradle folks.

src/main/groovy/org/labkey/gradle/plugin/UiTest.groovy
labkey-susanh: I shall attempt this.  It may mean that running the tests for an individual module doesn't work without also including these dependencies as well, but it's not something that's really supported or used now anyway.

labkey-tchad (approved):

PR #27, opened 2019-03-19
Fb 1.4.2 snapshot

labkey-susanh:


src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -37,27 +37,14 @@ class BuildUtils
     public static final String BUILD_FROM_SOURCE_PROP = "buildFromSource"
     public static final String BUILD_CLIENT_LIBS_FROM_SOURCE_PROP = "buildClientLibsFromSource"
     public static final String SERVER_MODULES_DIR = "server/modules"
+    public static final String PLATFORM_MODULES_DIR = "server/modules/platform"
+    public static final String COMMON_ASSAYS_MODULES_DIR = "server/modules/commonAssays"
```
labkey-tchad, line 41: This casing doesn't match the actual casing for the common assays repository. I don't know if it would actually affect anything but I'm wary of mismatched casing.

labkey-tchad (changes requested):
Just a question about directory casing

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-susanh: Thanks for catching that.  And BOOO! for not using camelCasing. @labkey-ians is the all-lowercase version set in stone?

labkey-susanh: Repository name has been updated.  Thanks, Ian!  Good to go, @labkey-tchad

labkey-tchad (approved):

PR #30, opened 2019-05-20
Version 1.5.0 

labkey-susanh:


build.gradle
```diff
@@ -52,7 +52,7 @@ buildscript {
 apply plugin: "org.ajoberstar.grgit"
 apply plugin: 'maven-publish'
 
-project.version = "1.4.4-SNAPSHOT"
+project.version = "1.4.5-SNAPSHOT"
```
labkey-tchad, line 55: Should this be 1.5.0? How do we decide which number to increment with a release?

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -557,20 +599,25 @@ class BuildUtils
...
+
+        if ('apiCompile'.equals(projectConfig))
+            return "${Api.CLASSIFIER}"
+        else if ('xmlSchema'.equals(projectConfig))
+            return "${XmlBeans.CLASSIFIER}"
+        else if ('jspCompile'.equals(projectConfig))
+            return "${Jsp.CLASSIFIER}"
+        // TODO when can the following be removed?
+        else if ('transformCompile'.equals(projectConfig)) // special business for CNPRC's distribution so it can include the genetics transform jar file
+            return "transform"
```
labkey-tchad, line 624: Seems like it could be generally useful in the future. Maybe just remove the note about CNPRC

labkey-tchad (approved):
Just wondering about the version number. I presume you just didn't want to confuse the git migration branch.
TeamCity is running this through its paces. Looking good so far.
https://teamcity.labkey.org/project.html?projectId=LabKey_GitMigration_Community

build.gradle
labkey-susanh: Yes, it will be 1.5.0, but using 1.4.5-SNAPSHOT here is still fine.  The snapshot is really just the best guess as to the next version.  We decide which number to increment based on semver rules.  In this case, since we've added some functionality (not just fixed existing functional) and changed no existing functionality, we'll increase the minor version.

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-susanh: Not sure it will be generally useful unless we have some intention to support a particular transformations.

src/main/groovy/org/labkey/gradle/task/RunTestSuite.groovy
```diff
@@ -42,8 +44,6 @@ class RunTestSuite extends RunUiTest
             dependsOn(project.project(':tools:Rpackages:install'))
         if (!project.getPlugins().hasPlugin(TeamCity.class) && project.tasks.findByName('packageChromeExtensions') != null)
             dependsOn(project.tasks.packageChromeExtensions)
-        if (project.findProject(":tools:Rpackages:install") != null)
-            dependsOn(project.project(':tools:Rpackages:install'))
```
labkey-susanh, line 46: @labkey-tchad  When did this dependency on Rpackages go away?

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -128,10 +128,13 @@ class BuildUtils
...
     {
-        settings.include ":sampledata:qc"
+        if (new File(rootDir, "sampledata/qc").exists())
+            settings.include ":sampledata:qc" // TODO: Remove when 19.2 is no longer supported
+        else
+            settings.include "${getTestProjectPath(settings.gradle)}:data:qc"
         settings.include getTestProjectPath(settings.gradle)
         includeModules(settings, rootDir, ["${convertPathToRelativeDir(getTestProjectPath(settings.gradle))}/modules"], [])
-        // TODO get rid of this when we decide whether to move dumbster
+        // TODO Remove when 19.2 is no longer supported. Dumbster moved into test modules
```
labkey-susanh, line 137: Shouldn't this say 19.1 is no longer supported?  We're moving this for 19.2

src/main/groovy/org/labkey/gradle/task/RunTestSuite.groovy
```diff
@@ -33,6 +34,7 @@ class RunTestSuite extends RunUiTest
         project.logger.info("RunTestSuite: constructor");
         scanForTestClasses = false
         include "org/labkey/test/Runner.class"
+        // TODO: Remove when 19.2 is no longer supported. Dependency is now declared in `qc/build.gradle`
         if (project.findProject(":sampledata:qc") != null)
             dependsOn(project.project(":sampledata:qc").tasks.jar)
```
labkey-susanh, line 39: I *think* it will be ok to have that dependency in the qc/build.gradle directory (since s comes before t for the current location and the qc project will be under the test project in the new location), but beware that this kind of dependency can cause order-of-operations problems.  If the project where the dependency is declared is evaluated before the project it depends on, the dependency won't get established.

src/main/groovy/org/labkey/gradle/task/RunTestSuite.groovy
```diff
@@ -33,6 +34,7 @@ class RunTestSuite extends RunUiTest
         project.logger.info("RunTestSuite: constructor");
         scanForTestClasses = false
         include "org/labkey/test/Runner.class"
+        // TODO: Remove when 19.2 is no longer supported. Dependency is now declared in `qc/build.gradle`
```
labkey-susanh, line 37: Shouldn't this say 19.1?  We're targeting this change for 19.2.

labkey-tchad: Wasn't sure about that. If we claim to support a particular version of LabKey, does that mean any pre-release/alpha version or just the final release?.

src/main/groovy/org/labkey/gradle/task/RunTestSuite.groovy
labkey-tchad: I ran into that problem with the `medImmune/build.gradle`, which adds a `dependsOn` to all of the test tasks. I changed it to use lazy task configuration and the problem cleared up. I used the same code in `qc/build.gradle` and it worked as expected

src/main/groovy/org/labkey/gradle/task/RunTestSuite.groovy
labkey-tchad: It's declared few lines earlier. I was removing the duplicate.

src/main/groovy/org/labkey/gradle/task/RunTestSuite.groovy
labkey-susanh: Just the final release.  Pre-releases are inherently unstable in my opinion.

PR #34, opened 2019-06-13
1.6 snapshot

labkey-susanh:
* Update tasks for yarn support
* Remove compile-time dependency on local tomcat installation


labkey-tchad:
Added `-PgradlePluginsVersion=1.6-SNAPSHOT` to trunk configuration to verify overnight.

labkey-tchad (approved):
Looks good on TeamCity

PR #36, opened 2019-06-25
1.6.2 snapshot

labkey-susanh:
Fix for symlinkNode task (referencing unknown property)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@ but also make certain assumptions that you may not want to impose on your module
 
 ## Release Notes
 
+### version 1.6.2
+*Release*: ???
+(Earliest compatible LabKey version 19.2)
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -9,6 +9,12 @@ but also make certain assumptions that you may not want to impose on your module
 
 ## Release Notes
 
+### version 1.6.2
+*Release*: ???
+(Earliest compatible LabKey version 19.2)
```
labkey-tchad: Does this change actually break compatibility with 19.1?

labkey-tchad (changes requested):

README.md
labkey-susanh: Ah, no.  Just a typo there.  Thanks!

labkey-tchad (approved):

PR #37, opened 2019-08-21
1.6.3 snapshot

labkey-susanh:
* Don't use yarn if there is a package-lock.json file.
* Add listNodeProjects task for showing which projects are using node in their build and which package manager
* Add links for both yarn and npm binaries
* [Issue 38198](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38198): Find the nearest vcs root to use for populating vcs properties to accommodate modules nested in git repositories
* Modify incubating multi-git tasks to account for git modules not in optionalModules and testAutomation repo
* Add more incubating multi-git tasks: gitPull, gitPush, and gitStatus


labkey-tchad (approved):
Looks alright to me. Going to let TeamCity churn on this for a while then manually verify the VCS info for module containers.

PR #39, opened 2019-10-08
Issue 38416: Publish pom files to Artifactory and build dependency graph correctly

ilikepies101:


src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
```diff
@@ -407,62 +407,55 @@ class FileModule implements Plugin<Project>
...
-                        libs(MavenPublication) { pub ->
-                            if (project.hasProperty('module'))
-                                pub.artifact(project.tasks.module)
-                            if (project.hasProperty('apiJar'))
+                        if (project.hasProperty('module'))
+                        {
+                            modules(MavenPublication) { pub ->
+                                    // can't produce more than one main artifact for a given project, so we can put the .module
+                                    // artifact in a different group, but we need to make a corresponding pom file.  For now,
+                                    // we leave the .module in the same group as the api jar.
```
labkey-susanh: Please update this comment to reflect current behavior.

src/main/groovy/org/labkey/gradle/plugin/extension/LabKeyExtension.groovy
```diff
@@ -91,6 +91,36 @@ class LabKeyExtension
             pomProperties.put("Description", description)
         pomProperties.put("License", "The Apache Software License, Version 2.0")
         pomProperties.put("LicenseURL", "http://www.apache.org/licenses/LICENSE-2.0.txt")
+        pomProperties.setProperty("scope", "compile")
```
labkey-susanh: This scope isn't always accurate.  For the module files, it's more of a runtime thing than a compile thing, so I would take this out of the base properties and set it in the ApiPomProperties and ModulePomProperties methods instead.

src/main/groovy/org/labkey/gradle/plugin/extension/LabKeyExtension.groovy
```diff
@@ -91,6 +91,36 @@ class LabKeyExtension
...
+        return pomProperties
+    }
+
+    static Properties getApiPomProperties(String artifactPrefix, String description)
+    {
+        Properties pomProperties = getBasePomProperties(artifactPrefix, description)
+        pomProperties.put("groupId", "org.labkey")
```
labkey-susanh: I'm contemplating whether we may want to change this group to be something like `org.labkey.api`.  I like it for consistency with the `org.labkey.module` pattern, and the next gradle plugins release will not work for earlier versions of labkey in any case, but it may cause problems that I'm not seeing at the moment.   Stay tuned.

src/main/groovy/org/labkey/gradle/plugin/extension/LabKeyExtension.groovy
```diff
@@ -91,6 +91,36 @@ class LabKeyExtension
...
+        pomProperties.setProperty("artifactCategory", "apiLib")
+        return pomProperties
+    }
+
+    static Properties getApiPomProperties(Project project)
+    {
+        return getApiPomProperties(project.name, project.description)
+
```
labkey-susanh:
```suggestion
```

src/main/groovy/org/labkey/gradle/plugin/extension/LabKeyExtension.groovy
```diff
@@ -91,6 +91,36 @@ class LabKeyExtension
...
+    }
+
+    static Properties getModulePomProperties(Project project)
+    {
+        return getModulePomProperties(project.name, project.description)
+
```
labkey-susanh:
```suggestion
```

src/main/groovy/org/labkey/gradle/task/PomFile.groovy
```diff
@@ -30,54 +29,38 @@ import org.labkey.gradle.util.BuildUtils
...
-                        }
-                        // FIXME it's possible to have external dependencies but no dependencies.
-                        // add in the dependencies from the external configuration as well
-                        def dependenciesNode = asNode().dependencies.first()
-                        project.configurations.api.allDependencies.each {
-                            def depNode = dependenciesNode.appendNode("dependency")
-                            depNode.appendNode("groupId", it.group)
-                            depNode.appendNode("artifactId", it.name)
-                            depNode.appendNode("version", it.version)
+                    def dependenciesNode = asNode().dependencies.first()
```
labkey-susanh: Does this work even if there are no dependencies declared?

src/main/groovy/org/labkey/gradle/task/PomFile.groovy
```diff
@@ -102,4 +85,39 @@ class PomFile extends DefaultTask
                 }
             }.writeTo(getPomFile())
     }
+
+    void removeDependencies(Node root)
+    {
+        if(isModulePom)
```
labkey-susanh:
```suggestion
        if (isModulePom)
```

src/main/groovy/org/labkey/gradle/task/PomFile.groovy
```diff
@@ -102,4 +85,39 @@ class PomFile extends DefaultTask
...
+        {
+            // Replace dependencies created by gradle
```
labkey-susanh: Can we be more descriptive of what dependencies are being removed?  (All the dependencies are created by gradle. :-).  These are the tomcat lib dependencies, right?

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -638,7 +637,8 @@ class BuildUtils
 
         String extensionString = extension == null ? "" : "@$extension"
 
-        return "org.labkey:${moduleName}${versionString}${extensionString}"
+        String prefix = extension.equals("module") ? "org.labkey.module" : "org.labkey"
```
labkey-susanh: `prefix` should be called `group` here.  It also seems like a good idea to make constants out of these two group names.

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -611,7 +609,8 @@ class BuildUtils
                 if (depVersion == null)
                     depVersion = depProject.version
             }
-            parentProject.dependencies.add(parentProjectConfig, getLabKeyArtifactName(parentProject, depProjectPath, depProjectConfig, depVersion, depExtension), closure)
+
+            parentProject.dependencies.add(parentProjectConfig, getLabKeyArtifactName(parentProject, depProjectPath, depProjectConfig, depVersion, depExtension), {transitive isTransitive})
```
labkey-susanh: Though we may not have a current usage of the closure parameter here (it was used by the now-defunct ms1 module, I believe), I think we don't want to ignore it.  If you can add the transitive setting to the closure provided (not sure of the syntax for that, or if it's possible), I would do that.  Otherwise, I would use the closure parameter if defined and otherwise pass the one with the transitive setting as you have here (with the thought being that if someone provides a closure they could reasonably expect to have to also define this transitive property in that closure).

labkey-susanh (changes requested):

labkey-susanh (commented):
@ilikepies101 We should update the `project.version` property in the build.gradle file on this branch so it is something other than `1.8.0-SNAPSHOT` so it does not interfere with the version being published from the 1.8.0-snapshot branch.  I suggest something like `1.8.0_issue38416-SNAPSHOT`.

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
ilikepies101: Good call. Updated

src/main/groovy/org/labkey/gradle/plugin/extension/LabKeyExtension.groovy
ilikepies101: 👀

src/main/groovy/org/labkey/gradle/task/PomFile.groovy
ilikepies101: It's been working because we've been adding an empty dependency node in `removeDependencies()` after replacing the tomcat dependencies. I changed it since if we didn't have any tomcat dependencies, we could potentially have issues

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
ilikepies101: I think I combined them as succinctly as I could but let me know if you think of a better way to do it.

src/main/groovy/org/labkey/gradle/task/PomFile.groovy
```diff
@@ -102,4 +90,39 @@ class PomFile extends DefaultTask
...
             }.writeTo(getPomFile())
     }
+
+    void removeDependencies(Node root)
+    {
+        if (isModulePom)
+        {
+            // Replace all dependencies automatically added by gradle (tomcat lib dependencies, etc.)
+            def dependenciesNode = new Node(null, 'dependencies')
+            root.get('dependencies')?.first()?.replaceNode(dependenciesNode)
```
labkey-susanh: This fails if there are no dependencies with the error ("Cannot access first() from an empty list").

src/main/groovy/org/labkey/gradle/plugin/LabKey.groovy
```diff
@@ -36,7 +37,7 @@ class LabKey implements Plugin<Project>
     @Override
     void apply(Project project)
     {
-        project.group = LABKEY_GROUP
+        project.group = API_GROUP
```
labkey-susanh: Since this represents the group for the entire project, I would probably leave it as `org.labkey` (which could also be defined as a constant).  I don't think it will show up anywhere outside of the properties for the project for now, but if accessed it makes more sense that it isn't api-specific.

labkey-susanh (changes requested):

labkey-susanh (approved):

PR #40, opened 2019-10-09
Issue 38553

ilikepies101:
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=38553

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
@@ -36,6 +36,8 @@ import org.labkey.gradle.util.GroupNames
 import org.labkey.gradle.util.PropertiesUtils
 
 import java.util.regex.Matcher
+import java.util.regex.Pattern
+
 /**
  * This class is used for building a LabKey file-based module, which contains only client-side code.
  * It also serves as a base class for the Java module classes.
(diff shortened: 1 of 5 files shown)


src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
```diff
@@ -168,10 +170,30 @@ class FileModule implements Plugin<Project>
                             String newLine = line
                             while (matcher.find())
                             {
-                                newLine = newLine.replace(matcher.group(), (String) project.lkModule.getPropertyValue(matcher.group(1), ""))
+                                if(PropertiesUtils.LABKEY_VERSION_PATTERN.matcher(line))
```
labkey-susanh: I think this doesn't need to be as specialized as this.  If we add in logic to get this property to the ModuleExtension.setModuleProperties method, then the logic here can stay the same.  And, this property value should be available from the project itself (`project.getProperty('labkeyVersion')`), so there shouldn't be a need to parse this file ourselves.

labkey-susanh (changes requested):

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
ilikepies101: Awesome! Didn't know about the labkey version already existing in the project object. Much simpler code now.

labkey-susanh (approved):

PR #41, opened 2019-10-11
Issue 38426: Git checkout recognizes forked repos

ilikepies101:
Forked repos are not checked out by default (https://help.github.com/en/articles/searching-in-forks). Update the query parameter to check out forked repo.

https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=38426


labkey-susanh (changes requested):
I think we actually want the default behavior to be including forks for all of the tasks in the plugin, and I'm not sure when we would want to exclude them, so I think we could just remove the includeForks property and always append `fork:true` to `queryString`.

labkey-susanh (approved):

PR #42, opened 2019-10-15
Issue 38681: Rename references from jspCompile to jspImplementation

ilikepies101:


labkey-susanh (changes requested):
These instances of `jspCompile` are completely internal to our plugin, so I think I would prefer to leave them as `jspCompile` since it makes more sense to describe what's going on (we're compiling the jsp files into Java and then into classes).

PR #43, opened 2019-10-15
Fb 1.8.0 snapshot

labkey-susanh:


ilikepies101 (approved):

README.md
```diff
@@ -9,7 +9,19 @@ but also make certain assumptions that you may not want to impose on your module
 
 ## Release Notes
 
-### veresion 1.7.0
+### version ???
```
ilikepies101: Should we change this version number?

ilikepies101 (changes requested):

README.md
labkey-susanh: This seems a reasonably clear version number to me. :-) (Yes, thanks for the reminder.)

labkey-tchad:
Verifying on TeamCity:
Build: https://teamcity.labkey.org/viewLog.html?buildId=835460
BVTs: https://teamcity.labkey.org/viewLog.html?buildId=835442
Installers: https://teamcity.labkey.org/viewLog.html?buildId=835625

labkey-tchad:
Rerunning installer after build fix: https://teamcity.labkey.org/viewLog.html?buildId=835921
Verifying alternate JAVA_HOME for tomcat: https://teamcity.labkey.org/viewLog.html?buildId=835980

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
```diff
@@ -405,38 +407,69 @@ class FileModule implements Plugin<Project>
...
+                            apiLib(MavenPublication) { pub ->
+                                pub.groupId = LabKeyExtension.API_GROUP
+                                pub.artifact(project.tasks.jar)
                             }
                         }
                     }
 
                     if (BuildUtils.shouldPublish(project))
                     {
                         project.artifactoryPublish {
```
labkey-tchad, line 452: Can/should we do lazy task configuration here? Something like:
```suggestion
                         project.tasks.named("artifactoryPublish").configure {
```

labkey-tchad:
Build, BVT, and Installers look good. Custom JAVA_HOME for Tomcat works (The run "failed" because I intentionally picked a JDK that we no longer support).
I've switched the TeamCity trunk project to run 1.8.0-SNAPSHOT by default to get more thorough coverage and make sure I'm not missing some corner case.
Any further changes should be verified in a feature branch before adding them to fb_1.8.0-snapshot.

labkey-tchad (approved):
Just one possible place to use lazy task configuration.

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
labkey-susanh: Probably a good idea, but not needed at the moment.  This is a task that only gets configured when the property is set to do publishing, so not a lot of savings in any case.

ilikepies101 (approved):

PR #46, opened 2019-11-21
includeWarArchive -- support building .war distribution artifact

labkey-matthewb:


src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -325,9 +335,94 @@ class ModuleDistribution extends DefaultTask
...
+                exclude(name: "bootstrap.jar")
+            }
+
+            // NOTE: if we want to enable running w/o LabKeyBootstrapClassLoader we have to do some of its work here (see ExplodedModule)
+            // NOTE: in particular some files need to be moved out of the module and into the WEB-INF directory
+
+            explodedModules.forEach({ File moduleDir ->
+                String moduleName = moduleDir.getName()
 
+                println("MODULE: ${moduleName}")
```
labkey-susanh: This println should probably be removed.  If you want some echoing to go on, it's nicer to use `project.logger.debug` or `project.logger.quiet` (the equivalent of println), but without more info here, it's unclear who is telling you about a module.

labkey-susanh (changes requested):

labkey-susanh (approved):

PR #49, opened 2020-02-09
Modify installer to support mode without tomcat

bbimber:
Hello,

We deploy LabKey to our cluster, where it runs in a mode without tomcat.  Years ago I wrote our deploy code to copy/paste a bunch of this script, since I was unable to use this script verbatim because we are not attempting to start/stop the tomcat service.  However, it is pretty easy to just add an argument to the official script to avoid start/stopping the tomcat server.  This PR adds that.

While extracting the LK tar and copying modules+ pipline_libs isnt hard, it would be slightly better to use the official LK script for this, if something ever changes in the future in that process.

Thank you for considering this.

### DIFF
--- a/distributionResources/manual-upgrade.sh
+++ b/distributionResources/manual-upgrade.sh
@@ -70,7 +70,7 @@ print_usage()
     echo ""
     echo "        --tomcat_lk: for LabKey Internal Hosted Server Installed Only"
     echo ""
-    echo "        --skip_tomcat: the script will not attempt to start/stop tomcat.  you must do this manually before/after running the script. This might be suitable for remote pipeline servers as well.
+    echo "        --skip_tomcat: the script will not attempt to start/stop tomcat.  You must do this manually before/after running the script.
     echo ""
 }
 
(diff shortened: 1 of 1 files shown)


distributionResources/manual-upgrade.sh
```diff
@@ -70,6 +70,8 @@ print_usage()
     echo ""
     echo "        --tomcat_lk: for LabKey Internal Hosted Server Installed Only"
     echo ""
+    echo "        --skip_tomcat: the script will not attempt to start/stop tomcat.  you must do this manually before/after running the script. This might be suitable for remote pipeline servers as well.
```
labkey-susanh:
```suggestion
    echo "        --skip_tomcat: the script will not attempt to start/stop tomcat.  You must do this manually before/after running the script.
```
The mention of a suggestion about remote pipeline servers might only serve to confuse, particularly if it is not a more definite recommendation.

labkey-susanh (approved):
Seems fine as minimal modification to the existing script.  One suggestion for updating the usage text, but no further review is required.

distributionResources/manual-upgrade.sh
bbimber: Sure, seems like a good idea

bbimber:
@labkey-susanh thanks - i added your suggestion.  i am not able to merge this - do you do this, or do we need another LK reviewer?

PR #50, opened 2020-02-11
module.xml properties updates to support versioning changes: "version…

labkey-adam:
…" -> "schemaVersion", "labkeyVersion" -> "releaseVersion", remove "consolidateScripts"


README.md
```diff
@@ -8,6 +8,15 @@ but also make certain assumptions that you may not want to impose on your module
 [LabKey documentation](https://www.labkey.org/Documentation/wiki-page.view?name=gradleModules) for more information.
 
 ## Release Notes
+### version x.y.z
+*Released*: xxx
+(Earliest compatible LabKey version: 20.3)
+
+* Rename "version" property to "schemaVersion" and support null value
+* Rename "labkeyVersion" property to "releaseVersion"
+* Remove "consolidateScripts" property
```
labkey-susanh: Can you qualify these comments with something like "Change module properties/module.xml writer"?

src/main/groovy/org/labkey/gradle/plugin/extension/ModuleExtension.groovy
```diff
@@ -126,21 +120,17 @@ class ModuleExtension
...
-        if (modProperties.getProperty("ManageVersion") == null)
+        if (modProperties.getProperty("SchemaVersion") == null)
         {
-            if (isExternalModule)
-                modProperties.setProperty("ManageVersion", "false")
+            if (modProperties.getProperty("Version") == null)
+                modProperties.setProperty("SchemaVersion", "")  // Spring binds this as setSchemaVersion(null), which is what we want
             else
-                modProperties.setProperty("ManageVersion", "true")
+                modProperties.setProperty("SchemaVersion", modProperties.getProperty("Version"))  // For backward compatibility with old modules TODO: Remove
```
labkey-susanh, line 133: Do we never use want to use the SchemaVersion value? (I assume that's better handled in the code itself, but a comment to that effect would be handy here.)

labkey-susanh (changes requested):

src/main/groovy/org/labkey/gradle/plugin/extension/ModuleExtension.groovy
labkey-adam: We do use it... if non-null, it gets copied into the template. This is the SchemaVersion == null (not specified in module.properties) case. Maybe I'm not understanding what you're asking for...

README.md
labkey-adam: Updated

src/main/groovy/org/labkey/gradle/plugin/extension/ModuleExtension.groovy
labkey-susanh: Ah, right.  Wasn't seeing the else part.

labkey-susanh (approved):

PR #53, opened 2020-03-03
1.10.2 snapshot

labkey-susanh:
Fixes needed for yarn conversion (and one deprecation fix)

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/ModuleResources.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/ModuleResources.groovy
@@ -66,7 +66,12 @@ class ModuleResources implements Plugin<Project>
                     }
                 }
         if (!LabKeyExtension.isDevMode(project))
-            project.tasks.processModuleResources.doLast(new GzipAction())
+        {
+            GzipAction zipAction = new GzipAction();
+            zipAction.extraExcludes = ["views/**"];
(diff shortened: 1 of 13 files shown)


cnathe (approved):

src/main/groovy/org/labkey/gradle/plugin/ModuleResources.groovy
```diff
@@ -66,7 +66,12 @@ class ModuleResources implements Plugin<Project>
                     }
                 }
         if (!LabKeyExtension.isDevMode(project))
-            project.tasks.processModuleResources.doLast(new GzipAction())
+        {
+            GzipAction zipAction = new GzipAction();
+            zipAction.extraExcludes = ["views/**"];
```
labkey-matthewb, line 71: Or restrict to web/** rather than exclude views?

labkey-tchad (approved):
TeamCity has been running happily with this for quite a while.

src/main/groovy/org/labkey/gradle/plugin/ModuleResources.groovy
labkey-susanh: I'll look at that.  It requires a bit more fiddling with the GzipAction that's used in a couple of other places, but I could possibly not use it here.

PR #54, opened 2020-03-17
Revise manual-install.sh script to remove old JAR files first

labkey-jony:
Updated the script so now it looks for the six JAR files and deletes them if they're found before doing the copy of the tomcat-lib JARs to the CATALINA_HOME/lib directory. This will ensure full backward compatibility between old versions of LabKey that use this script and new ones where the database JARs are in the labkeywebapp directory.

### DIFF
--- a/distributionResources/manual-upgrade.sh
+++ b/distributionResources/manual-upgrade.sh
@@ -349,94 +349,20 @@ if [ $? != 0 ]; then print_error; fi # exit if the last command failed
 #
 # Remove all existing LabKey JAR files from TOMCAT_HOME/lib directory, if found
 #
-
-LKJARFILE1="javax.activation.jar"
-LKJARFILE2="labkeyBootstrap.jar"
-LKJARFILE3="mysql.jar"
-LKJARFILE4="jtds.jar"
-LKJARFILE5="mail.jar"
-LKJARFILE6="postgresql.jar"
-
-echo "Checking Tomcat lib directory for old LabKey JAR files..."
-echo ""
-LKJAR1=$(ls $CATALINA_HOME/lib | grep -i $LKJARFILE1)
-if [ -n "$LKJAR1" ]
-then
-        echo "Deleting the $LKJAR1 file..."
-        rm $CATALINA_HOME/lib/$LKJAR1
-        echo ""
-        echo "Deleted..."
-        echo ""
-else
-        echo "The file $LKJARFILE1 does not exist, skipping..."
-        echo ""
-fi
-
-LKJAR2=$(ls $CATALINA_HOME/lib | grep -i $LKJARFILE2)
-if [ -n "$LKJAR2" ]
-then
-        echo "Deleting the $LKJAR2 file..."
-        rm $CATALINA_HOME/lib/$LKJAR2
-        echo ""
-        echo "Deleted..."
-        echo ""
-else
-        echo "The file $LKJARFILE2 does not exist, skipping..."
-        echo ""
-fi
-
-LKJAR3=$(ls $CATALINA_HOME/lib | grep -i $LKJARFILE3)
-if [ -n "$LKJAR3" ]
-then
-        echo "Deleting the $LKJAR3 file..."
-        rm $CATALINA_HOME/lib/$LKJAR3
-        echo ""
-        echo "Deleted..."
-        echo ""
-else
-        echo "The file $LKJARFILE3 does not exist, skipping..."
-        echo ""
-fi
-
-LKJAR4=$(ls $CATALINA_HOME/lib | grep -i $LKJARFILE4)
-if [ -n "$LKJAR4" ]
-then
-        echo "Deleting the $LKJAR4 file..."
-        rm $CATALINA_HOME/lib/$LKJAR4
-        echo ""
-        echo "Deleted..."
-        echo ""
-else
-        echo "The file $LKJARFILE4 does not exist, skipping..."
-        echo ""
-fi
-
-LKJAR5=$(ls $CATALINA_HOME/lib | grep -i $LKJARFILE5)
-if [ -n "$LKJAR5" ]
-then
-        echo "Deleting the $LKJAR5 file..."
-        rm $CATALINA_HOME/lib/$LKJAR5
-        echo ""
-        echo "Deleted..."
-        echo ""
-else
-        echo "The file $LKJARFILE5 does not exist, skipping..."
-        echo ""
-fi
-
-LKJAR6=$(ls $CATALINA_HOME/lib | grep -i $LKJARFILE6)
-if [ -n "$LKJAR6" ]
-then
-        echo "Deleting the $LKJAR6 file..."
-        rm $CATALINA_HOME/lib/$LKJAR6
-        echo ""
-        echo "Deleted..."
-        echo ""
-else
-        echo "The file $LKJARFILE6 does not exist, skipping..."
-        echo ""
-fi
-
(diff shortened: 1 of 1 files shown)


labkey-jeckels:
I'll defer to @labkey-adam and @labkey-susanh for the approval and timing of getting this merged.

labkey-adam (dismissed):
I'm no shell scripting expert, but isn't it possible to use a for loop here to avoid writing the same block of code six times?

labkey-jony:
It is. But I wanted to just brew up something that worked first before going down that road. Much of the original script code isn't as elegant as it could be.

distributionResources/manual-upgrade.sh
```diff
@@ -346,6 +346,97 @@ if [ $? != 0 ]; then print_error; fi # exit if the last command failed
...
+        echo "Deleting the $LKJAR6 file..."
+        rm $CATALINA_HOME/lib/$LKJAR6
+        echo ""
+        echo "Deleted..."
+        echo ""
+else
+        echo "The file $LKJARFILE6 does not exist, skipping..."
+        echo ""
+fi
+
```
labkey-susanh: Perhaps something like this.  

```suggestion
#
# Remove all existing LabKey JAR files from TOMCAT_HOME/lib directory, if found
#
jars=( "javax.activation.jar" "labkeyBootstrap.jar" "mysql.jar" "jtds.jar" "mail.jar" "postgresql.jar"); 
for jar in "${jars[@]}"
do
    file=$CATALINA_HOME/lib/$jar
    if [ -f $file ]
    then
        echo "Deleting the $file file..."
        rm $file;
        ret_value=$?
        if [ $ret_value -ne 0 ]; then
            echo "FAILED"
        fi
    fi
done
```
The -n test doesn't do much here, I don' think, but checking if the file exists (-f) should help prevent false failures.  And it's good to check if the deletion succeeded or not.  I don't think we want to say anything if the files don't exist so the non-legacy upgrade will work silently.

labkey-susanh (changes requested):

labkey-jony:
@labkey-susanh beat me to the punch, but this is what I came up with, which is close to what she came up with too.

```
#!/bin/sh

LKJARFILES=("javax.activation.jar" "labkeyBootstrap.jar" "mysql.jar" "jtds.jar" "mail.jar" "postgresql.jar")

echo "Checking Tomcat lib directory for old LabKey JAR files..."
echo ""

for i in "${LKJARFILES[@]}"
do
LKJAR=$(ls $CATALINA_HOME/lib | grep -i $i)
if [ -n "$LKJAR" ]
then
        echo "Deleting the $i file..."
        rm $CATALINA_HOME/lib/$i
        echo ""
        echo "Deleted..."
        echo ""
else
        echo "The file $i does not exist, skipping..."
        echo ""
fi
done
```

distributionResources/manual-upgrade.sh
```diff
@@ -346,6 +346,23 @@ if [ $? != 0 ]; then print_error; fi # exit if the last command failed
...
+for jar in "${jars[@]}"
+do
+    file=$CATALINA_HOME/lib/$jar
+    if [ -f $file ]
+    then
+        echo "Deleting the $file file..."
+        rm $file;
+        ret_value=$?
+        if [ $ret_value -ne 0 ]; then
+            echo "FAILED"
```
labkey-susanh, line 365: I think failure to remove these jar files is possibly fatal, right @labkey-adam?  So perhaps we want to use print_error either here or after the for loop if any of the deletions has failed, which will cause the upgrade to fail.

labkey-susanh (changes requested):

distributionResources/manual-upgrade.sh
labkey-adam: I don't know if I'd say truly fatal... they could upgrade and remove them manually afterwards. The server will be quite cranky at startup if it sees them, so admins will be well aware.

labkey-susanh: They won't know that it's going to be cranky about this failure, though, so I think we should at least warn them that they'll need to do this manually if something did not get deleted.

labkey-adam: I agree we should warn

labkey-jony: I don't think a warning is necessary, not in that exact language anyway. The point of this part of the script is to remove the old JAR files in order for the next part to happen: Adding the new JAR files.

Previously, we never removed old JAR files and instead the script would just do a forced copy of all six JAR files from tomcat-lib to the CATALINA_HOME/lib directory. This was fine the old way when we needed to have all six of those JARs going to CATALINA_HOME/lib. But since we're planning on putting the database JARs under the labkeywebapp directory, we need something that will always remove those old DB JAR files out of the CATALINA_HOME/lib during the upgrade process so it won't cause a conflict with the new DB JAR file location under labkeywebapp.

This is why my version here only indicates that a file was skipped because it doesn't exist (https://github.com/LabKey/gradlePlugin/pull/54#issuecomment-600381245), which is okay for the newer versions of LabKey that will be putting those JARs under labkeywebapp and if anyone uses this newer upgrade script on older binaries, it still will drop in the six JAR files to CATALINA_HOME/lib since the forced copy line is still in the script after it does the delete.

labkey-susanh: I would prefer if we add a warning if one of the legacy files that is meant to be deleted and not replaced cannot be deleted for some reason.  It's unlikely to happen, but could and it just seems better to provide as much information as we can.

Just a small modification to what is here:
```
jars=( "javax.activation.jar" "labkeyBootstrap.jar" "mysql.jar" "jtds.jar" "mail.jar" "postgresql.jar"); 

deleteFailed=0
for jar in "${jars[@]}"
do
    file=$CATALINA_HOME/lib/$jar
    if [ -f $file ]
    then
        echo "Deleting the $file file..."
        rm $file;
        retValue=$?
        if [ $retValue -ne 0 ]; then
            echo "FAILED"
            deleteFailed=1
        fi
    fi
done

if [ $deleteFailed -eq 1 ]; then
    echo "Unable to delete one or more legacy jar files from $CATALINA_HOME/lib.  You will need to remove these manually in order for the server to start cleanly."
fi
```

labkey-jony: That looks good to me. It is unlikely, but this definitely covers us if it ever does occur.

labkey-jony: I did a quick test with the revisions that Susan did.

If the files exist, it'll state that the files are being deleted.

If the files don't exist, it won't output anything, which is fine.

I tried to change the attributes on one of my JAR files to make it immutable and then ran the script. However, it didn't output a delete failed, but it gave me a prompt asking if I wanted to delete the file I had set to being immutable.

labkey-susanh: Add a `-f` to the `rm` command to get rid of that prompt (`rm -f $file`)

labkey-jony:
@labkey-susanh I've re-tested again with your additional recommendation and it looks good to me. I think it can be merged.

labkey-susanh (approved):

labkey-jony:
I tried using this on one of our instances and it is failing with the new code.

./manual-upgrade.sh: 353: ./LabKey20.3Beta-65305.27-OConnor-bin/manual-upgrade.sh: Syntax error: "(" unexpected

I'm not sure why this doesn't work on Ubuntu like it did on my Mac, but my old code that I originally submitted works.

PR #55, opened 2020-03-27
Issue 40014: Auto-delete JDBC drivers from CATALINA_HOME/lib during p…

labkey-adam:
…ickPg and pickMsSql

### DIFF
--- a/src/main/groovy/org/labkey/gradle/task/DoThenSetup.groovy
+++ b/src/main/groovy/org/labkey/gradle/task/DoThenSetup.groovy
@@ -217,6 +217,10 @@ class DoThenSetup extends DefaultTask
                         }
             }
 
+        new File("${project.tomcat.catalinaHome}/lib/jtds.jar").delete();
+        new File("${project.tomcat.catalinaHome}/lib/mysql.jar").delete();
+        new File("${project.tomcat.catalinaHome}/lib/postgresql.jar").delete();
+
         // Then copy them into the tomcat/lib directory
         project.ant.copy(
                 todir: "${project.tomcat.catalinaHome}/lib",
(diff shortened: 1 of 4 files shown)


src/main/groovy/org/labkey/gradle/task/DoThenSetup.groovy
```diff
@@ -217,6 +217,10 @@ class DoThenSetup extends DefaultTask
                         }
             }
 
+        new File("${project.tomcat.catalinaHome}/lib/jtds.jar").delete();
+        new File("${project.tomcat.catalinaHome}/lib/mysql.jar").delete();
+        new File("${project.tomcat.catalinaHome}/lib/postgresql.jar").delete();
```
labkey-susanh: I assume you've done this with the files not available and not seen it cause problems?  Since ServerDeploy also captures these (and other) jar files in the TOMCAT_LIB_UNVERSIONED_JARS list, we might do well to put this into a separate list we could iterate through:
```
public static final List<String> JDBC_JARS = ['jtds.jar', 'mysql.jar', 'postgres.jar'];
public static final List<String> TOMCAT_LIB_UNVERSIONED_JARS = JDBC_JARS + [...others...]
...
JDBC_JARS.each(String name -> {String name ->
           new File("${project.tomcat.catalinaHome}/lib/${name}").delete()
})
```

labkey-susanh (approved):
One suggestion.  Not necessary, but cleaner.

src/main/groovy/org/labkey/gradle/task/DoThenSetup.groovy
labkey-adam: Yes, seems to work fine whether those files exist or not.

labkey-adam: I like your change. Are you suggesting that the delete itself happens in ServerDeploy or in the same place in DoThenSetup?

labkey-adam: Change made... seems to work

labkey-susanh (approved):
Looks good.

PR #56, opened 2020-04-13
sh shell does not support arrays.  Use a function instead.

labkey-susanh:
**Rationale**
Original changes to the manual-upgrade.sh script assumed a shell that supports arrays, but this is not always the case.  The script has been modified to not use arrays and use a function instead for code reuse.  This is a fix we need for 20.3, so we patch from 1.10.3.

**Changes**
* [Issue 40184](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40184): manual-upgrade.sh script revisions for JAR file removal not working on Ubuntu

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,11 @@ but also make certain assumptions that you may not want to impose on your module
 
 ## Release Notes
 
+### version TBD
+*Released*: TBD
+(Earliest compatible LabKey version: 20.3)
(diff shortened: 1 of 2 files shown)


labkey-adam (approved):

labkey-jony:
FYI - I was able to test this successfully on client system.

PR #58, opened 2020-05-12
Gradle 1.12.0 snapshot

labkey-susanh:
**Changes**
* Eliminate duplication of jsp jars in war files
* Gradle 7 deprecation updates
* Disable loading module resources from source on TeamCity
* Restore use of module properties for generating pom files
* Add PomFileHelper utility to prepare for removal of deprecated maven plugin (in favor of maven-publish)
* Use project's group as a prefix to determine the group for api and module artifacts
* Make xsdDocZip  and jsDocZip tasks for publishing the doc files


src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -39,10 +39,10 @@ class BuildUtils
     public static final String SERVER_MODULES_DIR = "server/modules"
     public static final String PLATFORM_MODULES_DIR = "server/modules/platform"
     public static final String COMMON_ASSAYS_MODULES_DIR = "server/modules/commonAssays"
-    public static final String CUSTOM_MODULES_DIR = "server/customModules"
+    public static final String CUSTOM_MODULES_DIR = "server/customModules" // TODO remove after 20.6
```
labkey-adam: Can't this be removed now? This is not a real module place, AFAIK...

labkey-adam (approved):

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-susanh: Sure. Done

labkey-tchad (approved):

PR #66, opened 2020-05-22
Remove now-defunct references to GXT and GWT-DND

labkey-jeckels:
#### Rationale
We don't use GXT and GWT-DND in the product anymore, so the Gradle build shouldn't require they have properties set

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,9 @@ but also make certain assumptions that you may not want to impose on your module
 
 ## Release Notes
 
-### version X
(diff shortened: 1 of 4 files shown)


labkey-susanh (changes requested):
Please update the README adding a note about this change.  The version and date in that file can be left TBD.

labkey-susanh (approved):

labkey-jeckels:
Done!

PR #70, opened 2020-06-09
Issue 40602: Do not add projects to dedupe configuration dependency if they have buildFromSource=false

labkey-susanh:
#### Changes
[Issue 40602](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=40602): We can't exclude external dependencies using the dedupe configuration for projects that are not being built.  We accounted for the case that the project may not exist, but also need to check if the buildFromSource property has been set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,11 @@ but also make certain assumptions that you may not want to impose on your module
 [LabKey documentation](https://www.labkey.org/Documentation/wiki-page.view?name=gradleModules) for more information.
 
 ## Release Notes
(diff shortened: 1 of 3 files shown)


labkey-tchad:
Kicked off some builds/tests on 20.3 with this to verify.

labkey-tchad:
Trying again after publishing (copied branch to `fb_...` to have TeamCity pick it up)

labkey-tchad (approved):
Builds, tests, and installers work.

PR #76, opened 2020-07-13
Allow for use of `logj2.xml` file based on presence of the property log4j2Version

labkey-susanh:
#### Rationale
We are updating to log4j2 and need to copy its XML configuration file instead of `log4j.xml` when we have the appropriate properties.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1378

#### Changes
* Allow for use of `logj2.xml` file based on presence of the property `log4j2Version`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,11 @@ on how to do that, including how to develop and test locally and the versioning 
 
 ## Release Notes
 
+### version TBD
+*Released*: TBD
+(Earliest compatible LabKey version: 20.7)
+* Allow for use of `logj2.xml` file based on presence of the property `log4j2Version`.
+
 ### version 1.14.0
 *Released*: 8 July 2020
 (Earliest compatible LabKey version: 20.7)
--- a/build.gradle
+++ b/build.gradle
@@ -57,7 +57,7 @@ dependencies {
 }
 
 
-project.version = "1.15.0-SNAPSHOT"
+project.version = "1.15.0-log4j2-SNAPSHOT"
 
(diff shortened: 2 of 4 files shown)


labkey-tchad (approved):
Do we need this functionality long-term? I could just configure TeamCity to use the fb version to validate the Log4j 2 updates.

README.md
```diff
@@ -12,6 +12,11 @@ on how to do that, including how to develop and test locally and the versioning
 
 ## Release Notes
 
+### version TBD
+*Released*: TBD
+(Earliest compatible LabKey version: 20.7)
```
labkey-ankurj, line 17: 20.11?

labkey-ankurj (approved):

labkey-susanh:
> Do we need this functionality long-term? I could just configure TeamCity to use the fb version to validate the Log4j 2 updates.

I think we will need to use log4j2.xml going forward, but @labkey-ankurj correct me if I'm wrong there.

README.md
labkey-susanh: Still works for 20.7 since 20.7 doesn't have the log4j2Version property.

labkey-ankurj: ok

labkey-ankurj:
> > Do we need this functionality long-term? I could just configure TeamCity to use the fb version to validate the Log4j 2 updates.
> 
> I think we will need to use log4j2.xml going forward, but @labkey-ankurj correct me if I'm wrong there.

Yes, that is correct. New log4j picks up the config file having name log4j2.xml from the classpath.

PR #79, opened 2020-07-23
Wait for Tomcat to shutdown gracefully

labkey-tchad:
#### Rationale
`TeamCity.ensureShutdown` sometimes kills Tomcat before it is able to finish shutting down gracefully from the normal `stopTomcat` task. This is preventing shutdown listeners from logging JSP error counts, action stats, and query stats after some suites on TeamCity.

#### Changes
* Wait for Tomcat to shutdown gracefully before killing VM directly

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
@@ -37,6 +37,7 @@ import org.labkey.gradle.util.DatabaseProperties
 import org.labkey.gradle.util.GroupNames
 import org.labkey.gradle.util.PropertiesUtils
 
+import java.time.Duration
 import java.util.regex.Matcher
 
 /**
@@ -49,6 +50,7 @@ class TeamCity extends Tomcat
     private static final String TEST_CONFIGS_DIR = "configs/config-test"
     private static final String NLP_CONFIG_FILE = "nlpConfig.xml"
     private static final String PIPELINE_CONFIG_FILE =  "pipelineConfig.xml"
+    private static final Duration TOMCAT_SHUTDOWN_TIMEOUT = Duration.ofSeconds(15);
 
     private TeamCityExtension extension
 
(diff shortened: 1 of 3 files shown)


src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
```diff
@@ -301,11 +301,26 @@ class TeamCity extends Tomcat
...
 
-    private static void connect(AttachingConnector connector, int port) throws IllegalConnectorArgumentsException, IOException, IncompatibleThreadStateException
+    private static void threadDumpAndKill(AttachingConnector connector, int port) throws IllegalConnectorArgumentsException, IOException, IncompatibleThreadStateException
     {
         Map<String, Connector.Argument> arguments = connector.defaultArguments();
         arguments.get("hostname").setValue("localhost");
         arguments.get("port").setValue(Integer.toString(port));
+        long startTime = System.currentTimeMillis()
+        println("Waiting for graceful Tomcat shutdown.")
+        while (System.currentTimeMillis() - startTime > 15_000)
```
labkey-susanh: Don't you want `<` here?

labkey-susanh (changes requested):

src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
labkey-tchad: Indeed I do.

labkey-susanh (approved):

src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
```diff
@@ -301,73 +301,92 @@ class TeamCity extends Tomcat
...
+            vm = connector.attach(arguments)
+        }
+        catch (IOException ignore)
+        {
+            println("Unable to connect to VM at localhost:" + port + ", VM may already be shut down")
+            return
+        }
+
+        println("Waiting for graceful Tomcat shutdown.")
+        while (System.currentTimeMillis() - startTime < 15_000)
```
labkey-adam: Maybe create a constant for this (int seconds = 15) and reference it in the message below ("did not shutdown in __ seconds").

labkey-adam (approved):

PR #81, opened 2020-08-19
Simplify the dependencies and tasks for our test plugins and modify `deployModule` to also deploy dependencies

labkey-susanh:
#### Rationale
The `deployModule` task is useful as a targeted way to build and deploy a module, but when a module has dependencies on other modules and will not run without those modules present it only makes sense that `deployModule` would also deploy those dependencies.

We modify the dependencies and tasks for the test plugin to make them simpler and to fix a dependency declaration that makes the `moduleUiTests` tasks not work.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/433
* https://github.com/LabKey/platform/pull/1498

#### Changes
* Update default project paths for current reality
* Update `deployModule` task to also copy in the modules that are required
* Fix `undeployModule` so it actually does something
* Update logic for getting LabKey artifact name to account for testAutomation jar file
* Move dependency declarations out of TestRunner plugin in favor of `testAutomation/build.gradle` declarations 
* Remove `testJar` task in favor of a default jar task declared in `testAutomatioin/build.gradle`

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ on how to do that, including how to develop and test locally and the versioning 
 
 ### version TBD
 *Released*: TBD
-(Earliest compatible LabKey version: 20.7)
+(Earliest compatible LabKey version: 20.9)
 * Wait for Tomcat to shutdown gracefully before killing VM directly
 * Allow test server credentials to be customized on TeamCity
 * Update default project paths for current reality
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -12,11 +12,19 @@ on how to do that, including how to develop and test locally and the versioning
 
 ## Release Notes
 
-### version 1.15.1
+### version TBD
 *Released*: TBD
 (Earliest compatible LabKey version: 20.7)
```
labkey-tchad: Does this break compatibility with 20.7/8? Seems like the the testAutomation changes would; not sure about the product build changes.
```suggestion
(Earliest compatible LabKey version: 20.9)
```

labkey-tchad (approved):

PR #82, opened 2020-08-25
Remove additional reference to testJar task that no longer is being built

labkey-susanh:
#### Rationale
Need to remove all the references to the testJar task, even the ones only used by TeamCIty

#### Related Pull Requests
* #81 

#### Changes
* Replace `testJar` with just plain `jar`


labkey-tchad (approved):
Looking good on TeamCity

PR #91, opened 2020-10-09
Allow resource reloading on TeamCity

labkey-tchad:
#### Rationale
The module editor needs access to the modules source location. 

#### Changes
* Add TeamCity property to allow LabKey to load resources from module source

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,11 @@ on how to do that, including how to develop and test locally and the versioning 
 
 ## Release Notes
 
+### version 1.18.1
+*Release*: ??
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -12,6 +12,11 @@ on how to do that, including how to develop and test locally and the versioning
 
 ## Release Notes
 
+### version 1.18.1
```
labkey-susanh, line 15: This is probably going to be 1.19.0, merged with the other changes now in develop, but generally I just leave this as TBD until making the release.

labkey-susanh (approved):

PR #95, opened 2020-10-21
Fix distribution names for full Git enlistment

labkey-tchad:
#### Rationale
Now that we are all in Git, the distribution archives include the entire commit hash, which is too long and not very useful. Additionally, we don't quite handle the case where there is no vcs information available (branchId comes back as blank, not "none"). While I was in here, I removed some old code for handling alpha release branches, which we haven't used since 19.1.
I also changed this to reference TeamCity's branch information, since we rarely create feature branches in the 'distributions' or 'server' repositories. This will allow distributions built on feature branches to be named accordingly on TeamCity.

#### Changes
* Don't include the vcs revision in distribution file names
* Use TeamCity branch information in module versions
* Remove `-bin` from distribution archive names
* Clean up now redundant methods in `ModuleDistribution.groovy`

### DIFF
--- a/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
+++ b/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
@@ -353,103 +353,70 @@ class BuildUtils
 
     static String getVersionNumber(Project project)
     {
+        String version = project.labkeyVersion
+
         if (project.hasProperty("versioning"))
         {
             String branch = project.versioning.info.branchId
-            // When a git module is on the "master" branch in git, this corresponds to the alpha branch
-            // at the root and we want to use that version for consistency with the SVN modules
-            if (branch.equalsIgnoreCase("master"))
-                return project.rootProject.version
-            else if (["trunk", "develop", "none"].contains(branch) || branch.toLowerCase().matches("release.*-snapshot"))
-                return project.labkeyVersion
-            else
+            if (!["trunk", "develop", ""].contains(branch) && !branch.toLowerCase().matches("release.*-snapshot"))
             {
-                String currentVersion = project.labkeyVersion
-                return currentVersion.replace("-SNAPSHOT", "_${branch}-SNAPSHOT")
+                Matcher matcher = Pattern.compile(".*fb_(.+)").matcher(branch)
+                if (matcher.matches()) {
+                    branch = matcher.group(1) // Trim leading 'fb_' or 'XX.Y_fb_'
+                }
+                version = version.replace("-SNAPSHOT", "-${branch}-SNAPSHOT")
             }
         }
-        return project.labkeyVersion
+
+        return version
     }
 
     /**
      * Gets the versioning string to be inserted into distribution artifacts.  This string has a slightly different
      * format depending on the branch in which the distributions are created:
-     *     Trunk - 19.1-SNAPSHOT-62300.800 (<Current labkeyVersion>-<VCSRevision>.<BuildCounter>)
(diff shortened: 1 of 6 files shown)


src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -353,103 +353,70 @@ class BuildUtils
...
-            if (branch.equalsIgnoreCase("master"))
-                return project.rootProject.version
-            else if (["trunk", "develop", "none"].contains(branch) || branch.toLowerCase().matches("release.*-snapshot"))
-                return project.labkeyVersion
-            else
+            if (!["trunk", "develop", ""].contains(branch) && !branch.toLowerCase().matches("release.*-snapshot"))
             {
-                String currentVersion = project.labkeyVersion
-                return currentVersion.replace("-SNAPSHOT", "_${branch}-SNAPSHOT")
+                Matcher matcher = Pattern.compile(".*fb_(.+)").matcher(branch)
```
labkey-susanh, line 364: This pattern could be a little greedy, but it would require some rather strange naming of a feature branch so it should be fine.

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -353,103 +353,70 @@ class BuildUtils
...
         {
             String branch = project.versioning.info.branchId
-            // When a git module is on the "master" branch in git, this corresponds to the alpha branch
-            // at the root and we want to use that version for consistency with the SVN modules
-            if (branch.equalsIgnoreCase("master"))
-                return project.rootProject.version
-            else if (["trunk", "develop", "none"].contains(branch) || branch.toLowerCase().matches("release.*-snapshot"))
-                return project.labkeyVersion
-            else
+            if (!["trunk", "develop", ""].contains(branch) && !branch.toLowerCase().matches("release.*-snapshot"))
```
labkey-susanh: Are you sure "none" is not a possibility still?  We should probably also include "master" here.

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -353,103 +353,70 @@ class BuildUtils
...
-     *     Alpha - 19.1Alpha3-62239.3 (<labkey release version>Alpha<sprint number>-<VCSRevision>.<BuildCounter>)
-     *     Beta - 17.3Beta-47540.6 (<labkey release version>Beta-<VCSRevision>.<BuildNumber>)
+     *     Trunk/Develop - 20.11-SNAPSHOT-1500 (<Current labkeyVersion>-[<VCSRevision>.]<BuildNumber>)
+     *     Beta - 20.11Beta-6 (<labkey release version>Beta-[<VCSRevision>.]<BuildNumber>)
      *     (Beta means we are in a release branch, but have not yet released and updated from the snapshot version)
-     *     Release - 17.3-47926.26 (<labkey release version>-<VCSRevision>.<BuildNumber>)
+     *     Release - 20.11.0-1 (<Current labkeyVersion>-[<VCSRevision>.]<BuildNumber>)
      * See Issue 31165.
-     * @param project the distribution project
+     * @param project the distribution project. e.g. ':distributions:community'
```
labkey-susanh:
```suggestion
     * @param project the distribution project. e.g. project(':distributions:community')
```

labkey-susanh (changes requested):
Just one small change, I think.

labkey-susanh (approved):

PR #98, opened 2020-10-28
Configure tasks for using Gradle's build cache

labkey-susanh:
#### Rationale
Gradle's [build caching](https://docs.gradle.org/current/userguide/build_cache.html) can significantly improve build time.  There are a few types of standard tasks that Gradle caches by default, but other tasks need to be designated as cacheable and need to have their inputs and outputs curated to assure caching is possible.

We also switch away from using a custom location for the target of the `npm_setup` task as I believe the use of the build directory for these tasks causes up-to-date checks to always fail and thus the task always runs.  The default location for the task's output is the `.gradle` directory of the project, and, while I don't generally like the build to pollute the source directory, this does seem a fairly standard usage and can be easily designated as to be ignored by git.

#### Related Pull Requests
* 

#### Changes
* Designate certain tasks as cacheable for more efficient builds
* Use default locations, not in the build directory, for NpmRun tasks for more efficient builds
* Remove InstallRLabkey and InstallRuminex tasks as they are defined separately where used for tests


cnathe (approved):
Not sure I have the qualifications to review groovy code, but the paths for the inputs/outputs look right to me.

labkey-tchad (approved):
Looks good to me.

PR #107, opened 2020-12-07
Publish 1.22.0 and update snapshot version

labkey-susanh:
#### Rationale
Moving on to the next version

#### Related Pull Requests
#106 
#105


labkey-nicka (approved):
A release which will [live in infamy](https://en.wikipedia.org/wiki/Infamy_Speech).

PR #108, opened 2020-12-12
Update pickdb tasks to update embedded/application.properties

labkey-ankurj:
#### Rationale
This PR updates the gradle tasks like gradle build, pickdb and may be others to work with spring boot.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/21

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/Tomcat.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/Tomcat.groovy
@@ -16,21 +16,28 @@
 package org.labkey.gradle.plugin
 
 import org.gradle.api.DefaultTask
+import org.gradle.api.GradleException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.file.DeleteSpec
 import org.gradle.api.tasks.Delete
+import org.labkey.gradle.plugin.extension.LabKeyExtension
+import org.labkey.gradle.plugin.extension.ServerDeployExtension
 import org.labkey.gradle.plugin.extension.TomcatExtension
 import org.labkey.gradle.plugin.extension.UiTestExtension
 import org.labkey.gradle.task.StartTomcat
 import org.labkey.gradle.task.StopTomcat
+import org.labkey.gradle.util.BuildUtils
 import org.labkey.gradle.util.GroupNames
+import org.labkey.gradle.util.PropertiesUtils
 
 /**
  * Plugin for starting and stopping tomcat
  */
 class Tomcat implements Plugin<Project>
 {
+    private static final String EMBEDDED_LOG_FILE_NAME = "logs/embeddedTomcat.log"
+
     @Override
     void apply(Project project)
     {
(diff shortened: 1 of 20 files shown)


labkey-susanh (approved):
@labkey-ankurj I'll approve this since you cannot, but can you please review my changes in this branch?

src/main/groovy/org/labkey/gradle/plugin/Tomcat.groovy
```diff
@@ -65,6 +74,62 @@ class Tomcat implements Plugin<Project>
...
+                        File logFile = new File(ServerDeployExtension.getEmbeddedServerDeployDirectory(project), EMBEDDED_LOG_FILE_NAME)
+                        if (!logFile.getParentFile().exists())
+                            logFile.getParentFile().mkdirs()
+                        if (!logFile.exists())
+                            logFile.createNewFile()
+                        FileOutputStream outputStream = new FileOutputStream(logFile)
+                        def env = []
+                        env += "PATH=${ServerDeployExtension.getEmbeddedServerDeployDirectory(project)}/bin${File.pathSeparator}${System.getenv("PATH")}"
+                        if (TeamCityExtension.isOnTeamCity(project))
+                            env +="R_LIBS_USER=${System.getenv("R_LIBS_USER") != null ? System.getenv("R_LIBS_USER") : project.rootProject.file("sampledata/rlabkey")}"
```
labkey-tchad: We should pass through `R_LIBS_USER` regardless of running on TeamCity. Also, we don't need to default to '`sampledata/rlabkey`' anymore.
```suggestion
                        if (System.getenv("R_LIBS_USER") != null)
                            env +="R_LIBS_USER=${System.getenv("R_LIBS_USER")}"
```

src/main/groovy/org/labkey/gradle/plugin/Tomcat.groovy
```diff
@@ -65,6 +74,62 @@ class Tomcat implements Plugin<Project>
...
+                    def command = "curl -X POST $endpoint"
+                    task.logger.quiet("Sending command to $endpoint")
+                    def proc = command.execute()
+                    proc.waitFor()
+                    if (proc.exitValue() != 0)
+                        task.logger.warn("Shutdown command exited with non-zero status ${proc.exitValue()}.")
+                    else
+                        task.logger.quiet("Shutdown successful")
+                }
+        }
```
labkey-tchad, line 69: Will this work on Windows?

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -153,15 +171,17 @@ class ModuleDistribution extends DefaultTask
...
         {
             zipArchives()
+            if (shouldBuildEmbeddedArchive())
+                embeddedTomcatZipArchive()
         }
-        if (includeWarArchive)
-        {
-            warArchives()
-        }
+
```
labkey-tchad: Should we allow a distribution to have only the embedded archive?
```suggestion

        if (!includeZipArchive && !includeTarGZArchive && shouldBuildEmbeddedArchive())
        {
            embeddedTomcatTarArchive()
        }
```

labkey-tchad (approved):
Do we want to match our Java styling for curly brackets (new line for open brackets)? I noticedf some places that were changed that way but some new code has it the old way (no new line before open brackets).

src/main/groovy/org/labkey/gradle/plugin/Tomcat.groovy
labkey-susanh: It should if the `curl` command is available.

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
labkey-susanh: Possibly, but the choice of Tar archive here is sort of arbitrary (although I also wonder if we still need to support both tar and zip archives, but that's for a different disucssion).   The `includeEmbeddedArchive` property could be a string property where we provide 'tar.gz', 'zip' or both ('tar.gz,zip').  Perhaps that's better.

labkey-susanh (commented):
> Do we want to match our Java styling for curly brackets (new line for open brackets)? I noticedf some places that were changed that way but some new code has it the old way (no new line before open brackets).

I had been using Java styling at least for functions and for if blocks and the like as that seemed the (reasonable) default for Groovy, but now IntelliJ seems to think this other styling is the default way to go.  Since the default stying for closures in Groovy, which are quite prevalent, is rather unclear to me (even from a personal preference standpoint), I just can't get very excited about it either way.

PR #114, opened 2021-01-20
Add isOpenSource property to ModuleDistribution task to support packaging differently licensed libraries

labkey-susanh:
#### Rationale
Our open source distributions should include only open source versions of our dependencies.

#### Related Pull Requests
* 

#### Changes
* Add optional `isOpenSource` property to `ModuleDistribution` task
* Wire up dependencies for distribution task so it will include the proper api module based on the `isOpenSource` property

### DIFF
--- a/src/main/groovy/org/labkey/gradle/task/DoThenSetup.groovy
+++ b/src/main/groovy/org/labkey/gradle/task/DoThenSetup.groovy
@@ -156,7 +156,7 @@ class DoThenSetup extends DefaultTask
                             line = line.replace("#context.webAppLocation=", "context.webAppLocation=")
                             line = line.replace("#spring.devtools.restart.additional-paths=", "spring.devtools.restart.additional-paths=")
                         }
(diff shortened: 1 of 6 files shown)


src/main/groovy/org/labkey/gradle/task/DoThenSetup.groovy
```diff
@@ -156,7 +156,7 @@ class DoThenSetup extends DefaultTask
                             line = line.replace("#context.webAppLocation=", "context.webAppLocation=")
                             line = line.replace("#spring.devtools.restart.additional-paths=", "spring.devtools.restart.additional-paths=")
                         }
-                        if (databaseProperties.getProperty("extraJdbcDataSource"))
+                        if (databaseProperties.hasProperty("extraJdbcDataSource") && databaseProperties.getProperty("extraJdbcDataSource"))
```
labkey-tchad: I think `hasProperty` is what I meant
```suggestion
                        if (databaseProperties.hasProperty("extraJdbcDataSource"))
```

labkey-tchad (approved):

labkey-tchad (approved):

PR #121, opened 2021-03-18
Add methods to allow external dependencies to be declared with enough data to generate jars.txt files

labkey-susanh:
#### Rationale
Maintaining the `jars.txt` files is tedious.  Much of the data in this file is also maintained within the `build.gradle` files so it will make for more consistency and eliminate the need for generating and comparing to the `dependencies.txt` file if we can produce the `jars.txt` from the build.  This PR adds a utility method and updates the `WriteDependenciesFile` task to get us one step closer to not needing to maintain the `jars.txt` files.

The following assumptions were made:
* Since our jar files contain version numbers, we do not need to list the version number separately in the `jars.txt`.
* The `Component` column does not provide much value and can be removed.
* The `LabKey Dev` column does not provide much value and can be removed.
* All dependencies should provide license information, otherwise the build should fail
* All external dependencies will be declared using the current method or the new method but we do not need to support a combination of both. 

If any assumptions are wrong, we can adjust.  Just starting minimal.  I'm not sure if the auditors that have asked for this kind of data are more interested in seeing the name of the jar file or more build-like coordinates (e.g., org.apache.commons:commons-math3:3.4.1), but if this is more informative, we can use that instead.

An example usage:
```
   BuildUtils.addExternalDependency(
           project,
           new ExternalDependency(
                   "org.apache.commons:commons-math3:${commonsMath3Version}",
                   "Apache",
                   "http://commons.apache.org/math/",
                   "Apache 2.0",
                   "http://www.apache.org/licenses/LICENSE-2.0",
                   "Lightweight, self-contained mathematics and statistics components"
           )
   )
```
And the corresponding `jars.txt` file:
```
{table}
Filename|Source|License|Purpose
commons-math3-3.6.1.jar|{link:Apache|http://commons.apache.org/math/}|{link:Apache 2.0|http://www.apache.org/licenses/LICENSE-2.0}|Lightweight, self-contained mathematics and statistics components
{table}
```

There may need to be some iteration on this to handle some special cases of dependencies, particularly for the `platform` modules, but the basics should work.

While this feature is incubating or in transition, we will continue to produce the `dependencies.txt` file and the checks on the credits page will remain.  Once all modules have transitioned to the new method of declaring dependencies, we can remove the generation of `dependencies.txt` and the corresponding checks.

#### Changes
* Add `BuildUtils.addExternalDependency` and `BuildUtils.addExternalDependencies` methods for declaring external dependencies with data needed to produce the `jars.txt` file
* Modify `WriteDependenciesFile` to also write a new jars.txt file if one of the new methods has been used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,13 @@ If you are making changes to the plugins, please see the [internal docs](https:/
 on how to do that, including how to develop and test locally and the versioning information.
 
 ## Release Notes
+### TBD
+*Released*: TBD
+(Earliest compatible LabKey version: 21.3)
+* (incubating) Add ability to generate jars.txt from the build 
+  * Add `BuildUtils.addExternalDependency` method for declaring dependencies to be included in the `jars.txt` file
+  * Modify `WriteDependenciesFile` to also write a new jars.txt file if `addExternalDependency` has been used
+
 ### 1.25.1
 *Released*: 15 February 2021
 (Earliest compatible LabKey version: 21.3)
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ dependencies {
 }
 
 group 'org.labkey.build'
-project.version = "1.26.0-jarsTxt-SNAPSHOT"
+project.version = "1.26.0-SNAPSHOT"
 
 gradlePlugin {
     // TODO after transitioning to using these plugin ids, remove the properties files from resources/META-INF.gradle-plugins
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ dependencies {
 }
 
 group 'org.labkey.build'
-project.version = "1.26.0-SNAPSHOT"
+project.version = "1.26.0-jarsTxt-SNAPSHOT"
 
 gradlePlugin {
     // TODO after transitioning to using these plugin ids, remove the properties files from resources/META-INF.gradle-plugins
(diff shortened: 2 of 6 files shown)


labkey-jeckels (approved):
I like the direction this is going.

I'd retain the "Component" column as the JAR file names can be pretty terse and many times the "Purpose" column doesn't include the name of the library. I agree that dropping the developer field makes sense.

labkey-tchad (approved):

labkey-adam:
What happens in the case of the custom (non-Maven Central) jars we've pushed to Artifactory?

labkey-adam (approved):

labkey-susanh:
> What happens in the case of the custom (non-Maven Central) jars we've pushed to Artifactory?

Same as happens today.  The dependency is really added to the project in the same way.  This just provides the mechanism to add other data long with that dependency declaration (that would now be captured in the jars.txt file)

labkey-adam:
My curiosity was around what the generated jars.txt looks like in that case. Do we have coordinates, etc. like for those available via maven central?

labkey-susanh:
> My curiosity was around what the generated jars.txt looks like in that case. Do we have coordinates, etc. like for those available via maven central?

Yes, all the dependencies have coordinates (meaning, group:name:version values), whether stored in Maven central or not.

PR #124, opened 2021-05-21
Issue 43207: Don't create a jar file for file modules

labkey-susanh:
#### Rationale
If a file module contains a jar file, even if it's an empty jar file, it can't be used in the module editor (since that's interpreted as a module that contains java code).  With PR #122, some dependencies were added that inadvertently caused the jar task for be run for every file module.  This PR fixes that

#### Related Pull Requests
* #122 

#### Changes
* Don't create or configure the `populateExplodedLib` task for file modules

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ on how to do that, including how to develop and test locally and the versioning 
 ### TBD
 *Released*: TBD
 (Earliest compatible LabKey version: 21.3)
-* [Issue 43207](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43207): When building a file module, we don't want to create a jar file because that prevents the module from being used in the module editor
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -11,6 +11,11 @@ If you are making changes to the plugins, please see the [internal docs](https:/
 on how to do that, including how to develop and test locally and the versioning information.
 
 ## Release Notes
+### TBD
+*Released*: TBD
+(Earliest compatible LabKey version: 21.3)
+* [Issue 43207](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43207): When building a file module, we don't want to create a jar file because that prevents the module from being used in the module editor
```
RosalineP: Small quibble but I think it's "module loader" here, as that was where the issue was encountered at least

RosalineP (approved):
Looks good! Only had a small comment. Thanks for hopping on this Susan!

PR #128, opened 2021-06-29
Enable building embedded distributions from standalone modules

labkey-tchad:
#### Rationale
In order to build embedded distributions from standalone modules, they need to be able to pull the embedded jar from Artifactory instead of building it from `:server:embedded`. This adds an `embedded` configuration to the distribution plugin that references ':server:embedded:jar' or a published embedded jar as needed.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/99

#### Changes
* Enable building embedded distributions from standalone modules
* Exclude standard upgrade script and README from embedded distributions
* [Issue 43356: manual-upgrade.sh script fails to upgrade standalone installer](https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=43356)

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
@@ -79,7 +79,7 @@ class Distribution implements Plugin<Project>
                 {
                     embedded
                 }
-        project.configurations.distribution.setDescription("Artifacts for creating an embedded LabKey distribution (aka installer)")
+        project.configurations.embedded.setDescription("Artifacts for creating a LabKey distribution (aka installer) with Tomcat embedded in it")
 
         if (project.configurations.findByName("utilities") == null)
         {
--- a/src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
@@ -53,7 +53,7 @@ class Distribution implements Plugin<Project>
             project.evaluationDependsOn(BuildUtils.getServerProjectPath(project.gradle))
         }
         // we also depend on the jar task from the embedded project, if available
-        if (BuildUtils.useEmbeddedTomcat(project)) {
+        if (BuildUtils.useLocalEmbeddedTomcat(project)) {
             project.evaluationDependsOn(BuildUtils.getEmbeddedProjectPath(project.gradle))
         }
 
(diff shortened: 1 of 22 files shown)


src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
```diff
@@ -75,6 +75,12 @@ class Distribution implements Plugin<Project>
                 }
         project.configurations.distribution.setDescription("Artifacts of creating a LabKey distribution (aka installer)")
 
+        project.configurations
+                {
+                    embedded
+                }
+        project.configurations.distribution.setDescription("Artifacts for creating an embedded LabKey distribution (aka installer)")
```
labkey-susanh:
```suggestion
        project.configurations.embedded.setDescription("Artifacts for creating an LabKey distribution (aka installer) with Tomcat embedded in it")
```

src/main/groovy/org/labkey/gradle/task/StopTomcat.groovy
```diff
@@ -29,7 +29,7 @@ class StopTomcat extends DefaultTask
     @TaskAction
     void action()
     {
-        if (BuildUtils.useEmbeddedTomcat(project))
+        if (BuildUtils.useLocalEmbeddedTomcat(project))
```
labkey-susanh: Someone who has built a standalone distribution that includes the embedded jar and then uses the stopTomcat task will have problems here.  Do we have any documentation or help for this situation?

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -474,9 +470,9 @@ class ModuleDistribution extends DefaultTask
 
         ant.zip(destfile: getEmbeddedZipArchivePath()) {
             zipfileset(dir: project.buildDir, prefix: archiveName) { include(name: serverJarFile.getName()) }
-            zipfileset(dir: utilsDir.path, prefix: "${archiveName}/bin")
-            zipfileset(dir: project.buildDir, prefix: archiveName, filemode: 744){
-                include(name: "manual-upgrade.sh")
+
+            if (!simpleDistribution) {
+                zipfileset(dir: utilsDir.path, prefix: "${archiveName}/bin")
```
labkey-susanh, line 478: Are you sure you can get rid of the setting of the filemode for `manual-upgradle.sh`?  Are we no longer building installers from a Windows machine?  I think that's where this need arose.

labkey-tchad: This removes that script from embedded distributions altogether. I don't think it actually works for embedded distributions.

labkey-susanh: Got it.  Didn't see the larger context here.

labkey-susanh (approved):

PR #133, opened 2021-09-24
Support test properties template

labkey-tchad:
#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/870

#### Changes
* Support `test.properties.dist` file in `testAutomation` repository
* Allow DoThenSetup task to pull extra data source configuration from `config.properties`
* Don't deploy tomcat jars when deploying with embedded Tomcat

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
@@ -88,10 +88,20 @@ class UiTestExtension
             {
                 // Create test.properties file if necessary
                 def propertiesDistFile = project.project(BuildUtils.getTestProjectPath(project.gradle)).file(propertiesDistFileName)
-                FileUtils.copyFile(propertiesDistFile, propertiesFile)
+                if (propertiesDistFile.exists())
+                {
+                    FileUtils.copyFile(propertiesDistFile, propertiesFile)
+                }
             }
-            // read test.properties file
-            PropertiesUtils.readProperties(propertiesFile, this.config)
+            if (propertiesFile.exists())
+            {
+                // read test.properties file
+                PropertiesUtils.readProperties(propertiesFile, this.config)
+            }
+            else
+            {
+                project.logger.info("Unable to locate test properties files. Running with defaults.")
+            }
         }
         // if the test.properties file is not available, all properties will need to be provided via project properties
         for (String name : config.propertyNames())
--- a/src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
@@ -15,6 +15,7 @@
  */
 package org.labkey.gradle.plugin.extension
 
+import org.apache.commons.io.FileUtils
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.labkey.gradle.util.BuildUtils
@@ -23,7 +24,8 @@ import org.labkey.gradle.util.PropertiesUtils
 
 class UiTestExtension
 {
-    String propertiesFile = "test.properties"
+    final String propertiesFileName = "test.properties"
+    final String propertiesDistFileName = "test.properties.dist"
 
     private Properties config = null
     private Project project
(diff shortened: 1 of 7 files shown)


src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
```diff
@@ -80,8 +82,17 @@ class UiTestExtension
...
 
         if (project.findProject(BuildUtils.getTestProjectPath(project.gradle)) != null)
-        // read test.properties file
-            PropertiesUtils.readProperties(project.project(BuildUtils.getTestProjectPath(project.gradle)).file(propertiesFile), this.config)
+        {
+            def propertiesFile = project.project(BuildUtils.getTestProjectPath(project.gradle)).file(propertiesFileName)
+            if (!propertiesFile.exists())
+            {
+                // Create test.properties file if necessary
+                def propertiesDistFile = project.project(BuildUtils.getTestProjectPath(project.gradle)).file(propertiesDistFileName)
```
labkey-susanh: What is this file doesn't exist?

src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
```diff
@@ -23,7 +24,8 @@ import org.labkey.gradle.util.PropertiesUtils
 
 class UiTestExtension
 {
-    String propertiesFile = "test.properties"
+    final String propertiesFileName = "test.properties"
+    final String propertiesDistFileName = "test.properties.dist"
```
labkey-susanh: I'm confused about the purpose of the `.dist` file.  I had thought it was going to be a distribution definition, but apparently that's not it.  Is it a secondary set of properties?  Why do we need/want that?  Why do we believe this file will be there when the properties file won't be?  When will it be there?  A file rename may be in order, but at least a comment or two about when, why, and what this file is would be helpful.

src/main/groovy/org/labkey/gradle/task/DoThenSetup.groovy
```diff
@@ -94,7 +94,7 @@ class DoThenSetup extends DefaultTask
                             line = line.replace("@@ldapSyncConfig@@-->", "")
                             return line
                         }
-                        if (project.hasProperty("extraJdbcDataSource"))
+                        if (configProperties.hasProperty("extraJdbcDataSource"))
```
labkey-susanh, line 97: The handling of this is at least different than the other properties mentioned here (e.g., `enableLdbapSync`).  I'm guessing either this check was incorrect before or the shape of the TeamCity properties has changed.  Can we add some comments to the getExtraJdbcProperties method to indicate what these properties are expected to look like?

labkey-susanh (changes requested):

src/main/groovy/org/labkey/gradle/task/DoThenSetup.groovy
labkey-tchad: The check was incorrect before, and inconsistent between the embedded and non-embedded versions. I'm not sure who uses `enableLdapSync` and `enableJms` or when so I didn't want to touch them.
This change allows you to define a second datasource when deploying by specifying the jdbc properties in your `pg.properties` or `mssql.properties` files. Currrently, it assumes the properties will come if as Gradle properties (sufficient for TeamCity but kind of a pain when deploying locally).

src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
labkey-tchad: My intention is that the dist file is checked in to source control and folks don't touch it (similar to our intellij config templates). The existing `test.properties` will be deleted and added to `.gitignore` in testAutomation. `UiTestExtension` will generate `test.properties` automatically if it doesn't exist and folks can put their local configurations in there without cluttering their changelist.

src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
labkey-tchad: Added a check. It will now just pass through if the file is missing.

src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
labkey-susanh: I'm still not wild about the use of `dist` here just because it means something distinctly different elsewhere. I don't insist, but would `template` or `config` in place of `dist` be suitable?  In any case, comments on these two file names seem like they would be useful (e.g., for `propertiesDistFileName` it could be "the file checked in to VCS that is copied to propertiesFileName as part of the build" and for `propertiesFileName`, "the name of the file actually used in running tests").

labkey-susanh (approved):

src/main/groovy/org/labkey/gradle/plugin/extension/UiTestExtension.groovy
labkey-tchad: I was following some pattern I've see elsewhere (TeamCity comes with some default config files named with '.dist'). But you're correct, '.template' would be more consistent with elsewhere in LabKey; I'll make the switch.

PR #139, opened 2021-12-15
Use embedded config for declaring dependency on Springboot jar file

labkey-susanh:
#### Rationale
After the update to springBoot version 2.6, in conjunction with adding support for Java 16, the default jar file is now being created from the embedded project, and that default jar is what is being added to distributions.  We introduce a separate configuration that has the `bootJar` attached as an artifact and use that for declaring dependencies.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/169

#### Changes
* Use `embedded` configuration for declaring dependency when creating an embedded distribution
* Update to Gradle [7.3.2](https://github.com/gradle/gradle/releases/tag/v7.3.2)


labkey-tchad (approved):
Looking good on TeamCity: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Internal_Embedded_SampleManagerEmbeddedPostgres/1639318

gradlew
```diff
@@ -1,7 +1,7 @@
-#!/usr/bin/env sh
```
Sigmonia, line 1: Are there comparable changes for the Windows `.bat` script?

Sigmonia (approved):

gradlew
labkey-susanh: Apparently not.  I just did the standard `gradlew wrapper --gradle-version=7.3.2` update and took what changes arrived from that.

PR #144, opened 2022-02-11
Refine dependencies for SchemaCompile and other misc changes to avoid warnings

labkey-susanh:
#### Rationale
For some reason, perhaps because of the use of the ant builder, the build cache key for the `SchemaCompile` task does not change when the version of xmlBeans changes.  We add an input property for that version property to make it behave better.

We also take this opportunity to get rid of some warnings that show up  with Gradle 7.4 about deprecated properties and overlapping build directories.

#### Related Pull Requests
* 

#### Changes
* Update inputs of `SchemaCompile` task so it will defeat the cache when the xmlBeans version changes
* Update `RunUiTest` reports to non-deprecated properties
* Change `copyJspResources` to use `project.copy` instead of a `Copy` task to avoid warnings about cache optimizations that are not interesting

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,10 +15,10 @@ on how to do that, including how to develop and test locally and the versioning 
 
 _Note: 1.28.0 and later require Gradle 7_
 
-### TBD
-*Released*: TBD
+### 1.32.2
+*Released*: 11 February 2022
 (Earliest compatible LabKey version: 22.2)
-* Update inputs of `SchemaCompile` task so it will defeat the cache when the xmlBeans version changes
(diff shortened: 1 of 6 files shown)


labkey-adam (approved):

labkey-adam:
Should we replace the ant builder at some point?

labkey-susanh:
> Should we replace the ant builder at some point?

Perhaps.  I believe I tried it at one point and found it not entirely straightforward so stayed with the specific xmlbeans task/method that comes with `antbuilder` since it is still very much a first-class option within Gradle.

labkey-tchad (approved):

PR #155, opened 2022-08-24
Fix `Distribution.inheritDependencies` so excluded modules work when not building from source

labkey-susanh:
#### Rationale
When not building from source, we were failing to check the excluded modules property (since that is a list of project paths, not dependency coordinates).  This adds a method for getting the path from the dependency and using that to check if the dependency should be excluded.

#### Related Pull Requests
* 

#### Changes
* Do exclusion check for ModuleDependencies as well as ProjectDependencies


labkey-adam (approved):

labkey-tchad (changes requested):
Could you pull my change in from develop and republish?

labkey-tchad (approved):

PR #157, opened 2022-10-24
Fixes and updates for various dependency problems

labkey-susanh:
#### Rationale
The check for overlapping module names was using a static variable to hold the existing names. This variable persists between different runs of gradle commands, so if you've update the module names, it will still complain about an overlap until you start a new gradle daemon. Here we put the map for tracking overlaps into an extension that is recreated with each task run.

Finding all the transitive dependency requirements when creating a distribution to include, say, biologics, does not need to be a manual process since we know we need to include all the modules that are declared as module dependencies. Here we introduce logic to pick up the transitive dependencies from the `modules` configuration.

Now that all modules are converted to automatically generate a `jars.txt` file when needed, there is no need for the `dependencies.txt` file to be written.

#### Related Pull Requests

- https://github.com/LabKey/platform/pull/3778
- https://github.com/LabKey/testAutomation/pull/1276
- https://github.com/LabKey/server/pull/331
- https://github.com/LabKey/distributions/pull/327

#### Changes
* Don't use static variable when checking for modules with name overlaps
* When adding distribution dependencies, include all dependencies from the modules configuration for each project
* [Issue 46212](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=46212) Remove writing of dependencies.txt file


src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -548,6 +549,37 @@ class BuildUtils
...
+        Set<String> pathsAdded = new HashSet<>();
+        addTransitiveModuleDependencies(distributionProject, distributionProject.findProject(depProjectPath), config, pathsAdded)
+    }
+
+    private static void addTransitiveModuleDependencies(Project distributionProject, Project depProject, String config, Set<String> pathsAdded)
+    {
+        if (depProject == null)
+            return
+
+        distributionProject.evaluationDependsOn(depProject.getPath())
```
labkey-tchad: Will this work for modules that aren't currently enlisted?
Presuming that it would not, what if we added a new task that verified that no distributions are missing any depended on modules. This wouldn't make distribution creation any easier per-se but it would give us something to run on TeamCity to validate distribution PRs so that we wouldn't produce any invalid distributions.

Also, would this pull in transitive dependencies for projects that aren't intending to produce a functional distribution? e.g. when `makeDistribution = false` in `ModuleDistribution` task

src/main/groovy/org/labkey/gradle/task/WriteDependenciesFile.groovy
```diff
@@ -22,14 +22,10 @@ import org.gradle.api.artifacts.Configuration
...
 import org.labkey.gradle.plugin.extension.ModuleExtension
-import org.labkey.gradle.util.BuildUtils
 import org.labkey.gradle.util.ExternalDependency
 
 import java.nio.charset.StandardCharsets
 
-// This task can no longer be cacheable since we don't currently declare the jars.txt file as an output file.
-// Once there are no manually maintained jars.txt files, this can be a cacheable task
-//@CacheableTask
 class WriteDependenciesFile extends DefaultTask
```
labkey-adam, line 29: Should this be annotated as a `@CacheableTask` now per the previous comment?

labkey-adam (approved):

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-tchad: Answering my own question: yes, `discvr_modules` distribution includes a bunch of extra modules.

src/main/groovy/org/labkey/gradle/task/WriteDependenciesFile.groovy
labkey-susanh: I opted not to a this juncture since I'm not convinced it will do the right thing for all version number property changes.

labkey-adam: Okay

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-susanh: This should also work for modules that are not currently enlisted. The published pom file for a module includes all the modules it depends on, so they should get pulled in.  It could be that I'm missing a check for the existence of the project in the recursive case, but I think this generally works.  

We can certainly not do this adding of transitive dependencies if `makeDistribution = false`. Does that seem like the right thing to do? (Wasn't aware of/had forgotten that these non-distribution distributions existed.)

labkey-tchad: Isn't it too late to change the dependencies by the time the distribution task (and where `makeDistribution` is defined) is created?
What if we added a flag to `BuildUtils.addModuleDistributionDependencies` to control whether transitive dependencies should be included?

labkey-tchad: Also verified that this pulls in transitive module dependencies for modules that are not enlisted.

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
```diff
@@ -43,13 +43,13 @@ import java.util.regex.Matcher
...
 class FileModule implements Plugin<Project>
 {
-    private static final Map<String, String> _foundModules = new HashMap<>();
-
     @Override
     void apply(Project project)
     {
         def moduleKey = project.getName().toLowerCase()
-        def otherPath = _foundModules.get(moduleKey)
+        ServerDeployExtension deployExt = BuildUtils.getServerProject(project).extensions.getByType(ServerDeployExtension.class)
```
labkey-tchad: This will break standalone module builds (such as `tutorialModules/standalone`) since there isn't a `server` project.

labkey-tchad (changes requested):
Discovered a build break in standalone module builds.

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
labkey-susanh: Ah, good catch. I'll likely just disable this if there isn't a server project, since this is just a nicety to fail early so other things don't go haywire, right?

labkey-tchad: Correct.

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-adam: WRT `makeDistribution = false`, I'm not sure what it would mean to skip transitive dependencies when not making a distribution. Fairly sure we'd want distributions that inherit from these "non-distribution distributions" to inherit transitive dependencies as well.

labkey-tchad: non-distribution distributions are a different matter. Those are distributions without a `ModuleDistribution` task (e.g. `base`). The `makeDistribution` flag is to just zip up a collection of modules.

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
```diff
@@ -43,25 +43,25 @@ import java.util.regex.Matcher
...
+
+            if (otherPath != null && !otherPath.equals(project.getPath()) && project.findProject(otherPath) != null) {
+                if (shouldBuild)
+                    throw new IllegalStateException("Found duplicate module '${project.getName()}' in ${project.getPath()} and ${otherPath}. Modules should have unique names; Rename one or exclude it from your build.")
+            } else {
+                if (shouldBuild)
+                    deployExt.addFoundModule(moduleKey, project.getPath())
+                else
+                    deployExt.removeFoundModule(moduleKey)
+            }
```
labkey-tchad: I can't, for the life of me, remember why I had this remove the previously found module in the old implementation. Seems like a bug, but probably harmless.
```suggestion
            } else if (shouldBuild) {
                deployExt.addFoundModule(moduleKey, project.getPath())
            }
```
We could probably skip this whole duplicates check if `shouldBuild` is false (along with checking for the server project)

labkey-tchad (approved):

PR #159, opened 2022-12-07
Add tasks to generate startup properties files on TeamCity

labkey-tchad:
#### Rationale
In order to efficiently test different distributions or product tiers, we need a way to customize Tomcat startup more easily on TeamCity.

#### Related Pull Requests
* N/A

#### Changes
* Add `createStartupPropertyFile` task to define arbitrary LabKey startup properties
  * Controlled by `labkey.startup.properties` TeamCity property
* Add `includeDistModules` task to populate `ModuleLoader.include` startup property
  * Controlled by `labkey.startup.includeDistModules` TeamCity property

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,10 @@ _Note: 1.28.0 and later require Gradle 7_
 ### 1.36.2
 *Released*: TBD
 (Earliest compatible LabKey version: 22.9)
-* Add tasks to generate startup properties files on TeamCity
+* Add `createStartupPropertyFile` task to define arbitrary LabKey startup properties
+  * Controlled by `labkey.startup.properties` TeamCity property
+* Add `includeDistModules` task to populate `ModuleLoader.include` startup property
+  * Controlled by `labkey.startup.includeDistModules` TeamCity property
 
 ### 1.36.1
 *Released*: 18 November 2022
--- a/README.md
+++ b/README.md
@@ -15,8 +15,8 @@ on how to do that, including how to develop and test locally and the versioning 
 
 _Note: 1.28.0 and later require Gradle 7_
 
-### 1.36.2
-*Released*: TBD
+### 1.37.0
+*Released*: 8 December 2022
 (Earliest compatible LabKey version: 22.9)
 * Add `createStartupPropertyFile` task to define arbitrary LabKey startup properties
   * Controlled by `labkey.startup.properties` TeamCity property
--- a/README.md
+++ b/README.md
@@ -15,6 +15,11 @@ on how to do that, including how to develop and test locally and the versioning 
 
 _Note: 1.28.0 and later require Gradle 7_
 
+### 1.36.2
+*Released*: TBD
+(Earliest compatible LabKey version: 22.9)
+* Add tasks to generate startup properties files on TeamCity
+
 ### 1.36.1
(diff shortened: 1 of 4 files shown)


labkey-tchad:
Here's a build using the new parameters: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_ProductSuites_SampleManagerStarter_SampleManagerStarterPostgres/2158929?buildTab=artifacts#%2Fconfig%2Fstartup

README.md
```diff
@@ -15,6 +15,14 @@ on how to do that, including how to develop and test locally and the versioning
 
 _Note: 1.28.0 and later require Gradle 7_
 
+### 1.36.2
```
labkey-susanh: This will actually probably be version 1.37.0 (adding new functionality).

src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
```diff
@@ -237,6 +256,33 @@ class TeamCity extends Tomcat
...
+            project.evaluationDependsOn(inheritedDistPath)
+            def includeDistModulesTask = project.tasks.register("includeDistModules", Task) {
+                Task task ->
+                    task.group = GroupNames.TEST_SERVER
+                    task.description = "Generate server properties file to run with modules from a specified distribution"
+                    task.doLast {
+                        List<String> includeModules = new ArrayList<>();
+                        project.project(inheritedDistPath).configurations.distribution.dependencies.each {
+                            Dependency dep ->
+                                if (dep instanceof ProjectDependency || dep instanceof ModuleDependency) {
```
labkey-susanh: Is the if statement necessary? What are the other kinds of dependencies we're trying to avoid? (In any case, I ProjectDependency isa ModuleDependency, so one of these is not needed.)

src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
```diff
@@ -237,6 +256,33 @@ class TeamCity extends Tomcat
...
+                                    includeModules.add(dep.getName())
+                                }
+                        }
+                        extension.writeStartupProperties('00_modulesInclude.properties',
+                                'ModuleLoader.include;startup=' + String.join(',', includeModules))
```
labkey-susanh, line 274: Is this a startup property the exclusive set of properties to include or does it just include ones that might not otherwise be there?

labkey-susanh (approved):

src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
labkey-tchad: I suppose not. I adapted this from `Distribution.inheritDependencies` without fully understanding that the ifs there were just in service of excluding modules (which isn't an option here).

src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
labkey-tchad: From my experimentation, it seems to be the exclusive set of modules to include. `ModuleLoader.exclude` is applied afterward to remove some modules.

PR #160, opened 2022-12-09
Add @labkey/premium to list of package names for purging

labkey-susanh:
#### Rationale
We've added a new npm package so we want to purge versions of that package when doing cleanup

#### Changes
* Update list of packages to check versions of


cnathe (approved):
Thank you!

PR #162, opened 2022-12-30
Issue 43245: Update `ClientLibsCompress` task to use npm packages instead of the ancient YuiCompressor

labkey-susanh:
#### Rationale
From [Issue 43245](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=43245): For a long time now we've used YUI Compressor v2.4.8 (released 7/2/2013) to minify our JS and stylesheet assets declared in any LabKey `lib.xml` files. Over time this package has really started to show its age as it is falling further and further behind what modern browsers support. Often times this comes up as errors in our production build as a difficult to track down error when someone uses a valid browser supported JS/CSS syntax that YUI does not support.

The modern options for minification are available from client-side build technologies. One approach we might take is to simply require that modules with lib.xml files create package.json files that do the desired compression. I think this is not unreasonable as these files are generally not very volatile, but by continuing to do this automatically we can save some build time (installing the required `node_modules` only once) and not require that folks remember to update their package.json files when the lib.xml files are updated. 

The approach taken is as follows:
- There is a new Gradle project directory in the server build (`:server:minification`). This project contains a `package.json` file configured with the `devDependencies` required for doing the minification and the required `postcss.config.js` file that applies the `cssnano` plugin.  We could also locate a config file for `terser` here if the default doesn't do all we want it to.
- We apply the NpmRun plugin in`:server:minification` so we can use the npm executable that is installed by that plugin for executing the build scripts and we can install the node modules in this one directory for use in all subdirectories.  
- For each `lib.xml` file we find that references css or javascript files to be minified we create a directory under `server/minification/modules/<moduleName>` using the base name of the `lib.xml` file (e.g., `server/minification/modules/study/progress-report`). In this directory, we generate a `package.json` file with `minify-js` and/or `minify-css` scripts. The scripts reference the absolute paths of the input files and deposit the output files in the same directories we use for YuiCompression 
- The npm scripts are run for each directory

At the moment, you turn on the use of the minification via npm by adding the `useNpmMinifier` property to a project. It remains to be determined whether we need to retain the ability to fall back on the YuiCompressor, but before release of this version I will likely at least make the npm minifier the default and require a property such as `useYuiCompessor` to activate the other path. If we need to support the use of `lib.xml` files in standalone module builds, it will probably be easiest to retain the ability to use the YuiCompressor. However, my preference would be to guide people making standalone modules to set up a `package.json` file that does the require minification and doesn't rely on this special logic at all.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/379

#### Changes
* Update `ClientLibsCompress` task with option to use npm packages instead of the ancient YuiCompressor
  * Use [terser](https://github.com/terser/terser) for javascript minification 
  * Use [cssnano](https://cssnano.co/) via [postcss](https://github.com/postcss/postcss) for css minification

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,12 +15,13 @@ on how to do that, including how to develop and test locally and the versioning 
 
 _Note: 1.28.0 and later require Gradle 7_
 
-### TBD
-*Released*: TBD
+### 1.38.0
+*Released*: 9 January 2023
 (Earliest compatible LabKey version: 22.9)
-* Update `ClientLibsCompress` task to use npm packages instead of the ancient YuiCompressor
+* Update `ClientLibsCompress` task to use npm packages instead of the ancient YuiCompressor by default
   * Use [terser](https://github.com/terser/terser) for javascript minification 
   * Use [cssnano](https://cssnano.co/) via [postcss](https://github.com/postcss/postcss) for css minification
+  * Add check for `useYuiCompressor` property on projects to enable deprecated functionality
 
 ### 1.37.1
 *Released*: 9 December 2022
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ _Note: 1.28.0 and later require Gradle 7_
 (Earliest compatible LabKey version: 22.9)
 * Update `ClientLibsCompress` task to use npm packages instead of the ancient YuiCompressor
   * Use [terser](https://github.com/terser/terser) for javascript minification 
-  * Use [cssnano[(https://cssnano.co/)] via [postcss](https://github.com/postcss/postcss) for css minification
+  * Use [cssnano](https://cssnano.co/) via [postcss](https://github.com/postcss/postcss) for css minification
 
 ### 1.37.1
 *Released*: 9 December 2022
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@ _Note: 1.28.0 and later require Gradle 7_
 ### TBD
 *Released*: TBD
 (Earliest compatible LabKey version: 22.9)
-* Update `ClientLibsCompress` task to add option for not using ancient YuiCompressor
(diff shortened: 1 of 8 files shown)


cnathe (approved):

labkey-nicka (approved):

labkey-jeckels:
@labkey-susanh can you give more detail on what a "standalone module" means here? A module that people are developing without having other repos like `server` and `platform` checked out locally? Or would things like `ehrModules` qualify?

I think the set of external people who are building their own modules and care about minifying client libraries is exceedingly small, so I'm fine with supporting it via a `package.json` file but am trying to understand when it'll be needed.

labkey-susanh:
> @labkey-susanh can you give more detail on what a "standalone module" means here? A module that people are developing without having other repos like `server` and `platform` checked out locally? Or would things like `ehrModules` qualify?
> 
> I think the set of external people who are building their own modules and care about minifying client libraries is exceedingly small, so I'm fine with supporting it via a `package.json` file but am trying to understand when it'll be needed.

@labkey-jeckels a "standalone module" is one that is developed outside of the root `server` enlistment (and thus would not have the new `server/minification` project).

labkey-jeckels:
> > @labkey-susanh can you give more detail on what a "standalone module" means here? A module that people are developing without having other repos like `server` and `platform` checked out locally? Or would things like `ehrModules` qualify?
> > I think the set of external people who are building their own modules and care about minifying client libraries is exceedingly small, so I'm fine with supporting it via a `package.json` file but am trying to understand when it'll be needed.
> 
> @labkey-jeckels a "standalone module" is one that is developed outside of the root `server` enlistment (and thus would not have the new `server/minification` project).

Great. I'm fully on board with the plan for requiring `package.json` in that scenario.

PR #164, opened 2023-01-11
Correctly disable `npm` minifier if minification project is missing

labkey-tchad:
#### Rationale
`project.project()` throws an exception if the project isn't defined. Need to check for project definition as well as the directory.

#### Related Pull Requests
* #162 

#### Changes
* Correctly disable `npm` minifier if minification project is missing

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,8 @@ _Note: 1.28.0 and later require Gradle 7_
 
 ### 1.40.0-SNAPSHOT
 *Released*: TBD
+(Earliest compatible LabKey version: 22.9)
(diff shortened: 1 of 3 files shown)


labkey-tchad:
This fixes the `communityArtifacts` and `starterArtifacts` module builds (which don't include `:server:minification`). I presume this will also fix other standalone module builds once they update to a newer gradle plugin version.

labkey-susanh (approved):
Seems fine, but we'll also soon be removing the ability to use anything other than npm, so this method will go away.  For standalone modules, our recommendation will be that those modules use npm directly.

PR #168, opened 2023-03-03
Gradle 8.0.1

labkey-susanh:
#### Rationale
Gradle 8.0 is more restrictive about dependency declarations, insisting that you declare task dependencies when one task uses the output of another. TeamCity applies the artifactory plugin for publishing distributions to Artifactory, so we need to declare an explicit dependency between the `artifactoryPublish` task, if it exists, and the `distribution` task.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/408
* https://github.com/LabKey/server/pull/410

#### Changes
* Add explicit dependency declaration between artifactoryPublish task provided through TeamCity and distributions task
* Update some dependency versions
* Update gradle version to 8.0.1


src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
```diff
@@ -104,6 +104,9 @@ class Distribution implements Plugin<Project>
         {
             BuildUtils.addLabKeyDependency(project: project, config: "embedded", depProjectPath: BuildUtils.getEmbeddedProjectPath(project.gradle), depVersion: project.labkeyVersion, depProjectConfig: "embedded", transitive: false)
         }
+        if (project.tasks.findByName('artifactoryDeploy') != null) {
+            project.tasks.artifactoryDeploy.dependsOn(project.tasks.distribution)
+        }
```
labkey-tchad: I added a helper for this pattern at some point:
```suggestion
        TaskUtils.configureTaskIfPresent(project, 'artifactoryDeploy', { dependsOn(project.tasks.distribution) })
```

labkey-tchad (approved):

gradle.properties
```diff
@@ -7,10 +7,10 @@ commonsLang3Version=3.12.0
...
 grgitGradleVersion=5.0.0
-graphqlJavaVersion=18.1
+graphqlJavaVersion=20.0
 
 httpclientVersion=4.5.14
 httpcoreVersion=4.4.16
 
-jacksonVersion=2.13.3
-jsonVersion=20220320
+jacksonVersion=2.14.1
```
labkey-adam, line 15: `2.14.2` is available

labkey-susanh: Choosing 2.14.1 to align with sever build (for easier property reuse at the moment).

labkey-adam: Fair enough. Tangentially related: I've been thinking about adding a `commonsTextVersion` to the server `gradle.properties`. This library is only used in a couple EHR modules right now but 1) might as well sync them to the same version and 2) might actually be time to pull this into api since some commons lang methods we use are deprecated in favor of commons text. Not saying we need to do anything in this PR...

labkey-adam: Oh, cool, just saw your commit over in server.

labkey-susanh: I added `commonsTextVersion` over on server with a comment about its being used for gradle plugins. We can remove that comment if it's actually used elsewhere, but I'll leave it as is for now.

labkey-adam: Sounds good... thanks!

PR #169, opened 2023-03-15
Issue 47465: Check for the presence of the minification project 

labkey-susanh:
#### Rationale
The switch to using npm instead of yui-compressor for our npm and css minification introduced the :server:minification project that is used to do the work of minification. For a standalone build, this project won't exist. It is assumed standalone builds will do their own minification, if necessary, and likely won't have `.lib.xml` files in any case, so we simply won't try to add the tasks related to minification if the required project is not available.

#### Related Pull Requests
* https://github.com/LabKey/server/pull/417
* https://github.com/LabKey/tutorialModules/pull/113

#### Changes
* Add checks for the minification project before adding tasks and dependencies that rely on it.

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
@@ -207,7 +207,7 @@ class FileModule implements Plugin<Project>
 
             Task moduleFile = project.tasks.module
 
-            boolean haveMinifyProject = project.findProject(BuildUtils.getMinificationProjectPath(project.gradle)) != null
+            boolean haveMinifyProject = BuildUtils.haveMinificationProject(project.gradle)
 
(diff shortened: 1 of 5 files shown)


labkey-jeckels (approved):

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
labkey-tchad, line 1: Need to conditionalize the `compressClientLibs` dependency also

labkey-tchad (changes requested):

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
labkey-susanh: Thanks. Done.

labkey-tchad (approved):

PR #173, opened 2023-05-11
Improve audit trail of LabKey install script

labkey-tchad:
#### Rationale
[Issue 47854: Improve audit trail of LabKey install script](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=47854)

#### Related Pull Requests
* N/A

#### Changes
* Use `$(pwd)` to record distribution directory instead of `.`


labkey-susanh (approved):
Do we have a good way to test this?

I'll roll this into the next gradle plugins release, which will also include my change for the jsp configurations.

labkey-tchad:
> Do we have a good way to test this?
> 
> I'll roll this into the next gradle plugins release, which will also include my change for the jsp configurations.

Not really. Seems fairly low risk though. The script doesn't use this variable much. There's one spot that `cd`s to it but that's the only `cd` in the entire script; everything else is just validating or logging it.

PR #179, opened 2023-07-17
Set DuplicateStrategy for various copy tasks

labkey-susanh:
#### Rationale
Gradle wants you to explicitly tell it what DuplicatesStrategy to use for copy tasks. The `deployModule` task in particular was failing for me in one circumstance (not building some modules from source) for lack of a `DuplicateStrategy`. Probably the settings for tasks that copy into the build directory are superfluous, but setting to `INCLUDE` seems like the best default strategy in these cases

#### Related Pull Requests
* 

#### Changes
* Set `DuplicatesStrategy` explicitly for copy task
* Update graphqlVersion
* Gradle 8.2.1


src/main/groovy/org/labkey/gradle/task/DeployAppBase.groovy
```diff
@@ -54,6 +54,7 @@ class DeployAppBase extends DefaultTask {
                 project.copy { CopySpec copy ->
                     copy.from dir
                     copy.into deployBinDir.getPath()
+                    copy.setDuplicatesStrategy(DuplicatesStrategy.EXCLUDE)
```
labkey-susanh, line 57: I'm leaving this as EXCLUDE since that's what's used in the `if` statement above this copy (used only on TC).

labkey-adam (approved):

PR #185, opened 2023-11-14
Replace `buildDir` with `layout.buildDirectory` and do some other housekeeping

labkey-susanh:
#### Rationale
In Gradle 8.3, the use of `project.buildDir` was [deprecated](https://docs.gradle.org/current/userguide/upgrading_version_8.html#project_builddir) in favor of `project.layout.buildDirectory`.  We make adjustments here to upgrade to the new world order.  We also take this opportunity to do a little housecleaning of a good number of stray semicolons that are cluttering the space and to replace almost all remaining usages of `task.findTaskByName` with `task.named`.

#### Related Pull Requests
* 

#### Changes
* [Issue 49045](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49045) - make copying to the modules-api directory a proper task, so it can run even if apiJar does not
* Eliminate usages of deprecated `project.buildDir`
* Update `BuildUtils.includeModules` to be able to specify full module paths for exclusion, not just directory names

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/AntBuild.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/AntBuild.groovy
@@ -61,8 +61,8 @@ class AntBuild implements Plugin<Project>
     {
         project.ant.setProperty('basedir', BuildUtils.getServerProject(project).projectDir)
         project.ant.setProperty('modules.dir', project.projectDir.parent)
-        project.ant.setProperty('build.modules.dir', project.layout.buildDirectory.getAsFile().get().parent)
-        project.ant.setProperty('build.dir', project.rootProject.layout.buildDirectory.getAsFile().path)
+        project.ant.setProperty('build.modules.dir', BuildUtils.getBuildDir(project).parent)
+        project.ant.setProperty('build.dir', BuildUtils.getBuildDirPath(project))
         project.ant.setProperty('explodedModuleDir', project.labkey.explodedModuleDir)
         project.ant.setProperty('java.source.and.target', project.sourceCompatibility)
     }
--- a/src/main/groovy/org/labkey/gradle/plugin/AntBuild.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/AntBuild.groovy
@@ -61,8 +61,8 @@ class AntBuild implements Plugin<Project>
     {
         project.ant.setProperty('basedir', BuildUtils.getServerProject(project).projectDir)
         project.ant.setProperty('modules.dir', project.projectDir.parent)
-        project.ant.setProperty('build.modules.dir', project.buildDir.parent)
-        project.ant.setProperty('build.dir', project.rootProject.buildDir)
+        project.ant.setProperty('build.modules.dir', project.layout.buildDirectory.getAsFile().get().parent)
+        project.ant.setProperty('build.dir', project.rootProject.layout.buildDirectory.getAsFile().path)
         project.ant.setProperty('explodedModuleDir', project.labkey.explodedModuleDir)
         project.ant.setProperty('java.source.and.target', project.sourceCompatibility)
     }
--- a/src/main/groovy/org/labkey/gradle/plugin/AntBuild.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/AntBuild.groovy
@@ -62,7 +62,7 @@ class AntBuild implements Plugin<Project>
         project.ant.setProperty('basedir', BuildUtils.getServerProject(project).projectDir)
         project.ant.setProperty('modules.dir', project.projectDir.parent)
         project.ant.setProperty('build.modules.dir', BuildUtils.getBuildDir(project).parent)
-        project.ant.setProperty('build.dir', BuildUtils.getBuildDirPath(project))
+        project.ant.setProperty('build.dir', BuildUtils.getBuildDirPath(project.rootProject))
         project.ant.setProperty('explodedModuleDir', project.labkey.explodedModuleDir)
         project.ant.setProperty('java.source.and.target', project.sourceCompatibility)
     }
--- a/src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
@@ -18,7 +18,6 @@ package org.labkey.gradle.plugin
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.Task
-import org.gradle.api.UnknownTaskException
 import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.ModuleDependency
 import org.gradle.api.artifacts.ProjectDependency
@@ -142,9 +141,7 @@ class Distribution implements Plugin<Project>
                     if (it instanceof DefaultProjectDependency)
                     {
                         DefaultProjectDependency dep = (DefaultProjectDependency) it
-                        try {
-                            distTask.dependsOn(dep.dependencyProject.tasks.named("module"))
-                        } catch (UnknownTaskException ignore) {}
+                        TaskUtils.addOptionalTaskDependency(dep.dependencyProject, distTask, "module")
                     }
                 }
             }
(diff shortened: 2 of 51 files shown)


src/main/groovy/org/labkey/gradle/plugin/AntBuild.groovy
```diff
@@ -61,8 +61,8 @@ class AntBuild implements Plugin<Project>
     {
         project.ant.setProperty('basedir', BuildUtils.getServerProject(project).projectDir)
         project.ant.setProperty('modules.dir', project.projectDir.parent)
-        project.ant.setProperty('build.modules.dir', project.buildDir.parent)
-        project.ant.setProperty('build.dir', project.rootProject.buildDir)
+        project.ant.setProperty('build.modules.dir', BuildUtils.getBuildDir(project).parent)
+        project.ant.setProperty('build.dir', BuildUtils.getBuildDirPath(project))
```
labkey-tchad: Should this be for rootProject?
```suggestion
        project.ant.setProperty('build.dir', BuildUtils.getBuildDirPath(project.rootProject))
```

src/main/groovy/org/labkey/gradle/plugin/JavaModule.groovy
```diff
@@ -185,43 +191,39 @@ class JavaModule implements Plugin<Project>
...
-                {
-                    dependsOn(project.tasks.apiJar)
-                }
-                if (project.tasks.findByName('jspJar') != null)
-                {
-                    dependsOn(project.tasks.jspJar)
-                }
+                try {
+                    dependsOn(project.tasks.named("jar"))
+                } catch (UnknownTaskException ignore) {}
```
labkey-tchad: Could this use the new `TaskUtils` method?
```suggestion
                TaskUtils.addOptionalTaskDependency(project, it, "jar")
```

src/main/groovy/org/labkey/gradle/util/TaskUtils.groovy
labkey-tchad, line 1: Maybe add some more general-purpose methods to avoid more `UnknownTaskException` try-catch blocks.
```
    static TaskProvider doIfTaskPresent(Project project, String taskName, Consumer<TaskProvider> consumer)
    {
        return getFromTaskIfPresent(project, taskName, task -> {
            consumer.accept(task)
            return task
        })
    }

    static <R> R getFromTaskIfPresent(Project project, String taskName, Function<TaskProvider, R> func)
    {
        return getFromTaskIfPresent(project, taskName, func, null)
    }

    static <R> R getFromTaskIfPresent(Project project, String taskName, Function<TaskProvider, R> func, R defaultValue)
    {
        try {
            var task = project.tasks.named(taskName)
            return func.apply(task)
        }
        catch (UnknownTaskException ignore) {
            return defaultValue
        }
    }
```

src/main/groovy/org/labkey/gradle/plugin/Gwt.groovy
```diff
@@ -138,11 +139,11 @@ class Gwt implements Plugin<Project>
...
                         java.args =
                                 [
-                                        '-war', outputDir,
+                                        '-war', outputDirPath,
                                         '-style', project.gwt.style,
                                         '-logLevel', project.gwt.logLevel,
-                                        '-extra', extrasDir,
-                                        '-deploy', extrasDir,
+                                        '-extra', extrasDirPath,
+                                        '-deploy', extrasDirPath,
```
labkey-tchad, line 146: These don't seem to be defined anywhere. Should they be `outputDir.path` and `extrasDir.path`?

labkey-tchad (commented):
Submitting a partial review because some of these path differences seem like they might break things.

src/main/groovy/org/labkey/gradle/plugin/AntBuild.groovy
labkey-susanh: Yes, Thanks.

src/main/groovy/org/labkey/gradle/plugin/JavaModule.groovy
labkey-susanh: Yes, thanks.

src/main/groovy/org/labkey/gradle/plugin/Gwt.groovy
labkey-susanh: These are defined on lines 108 and 109.

src/main/groovy/org/labkey/gradle/util/TaskUtils.groovy
labkey-susanh: Sure. That does seem cleaner.

src/main/groovy/org/labkey/gradle/util/TaskUtils.groovy
```diff
@@ -27,4 +32,36 @@ class TaskUtils
...
+    static <R> R getFromTaskIfPresent(Project project, String taskName, Function<TaskProvider, R> func, R defaultValue)
+    {
+        try {
+            var task = project.tasks.named(taskName)
+            return func.apply(task)
+        }
+        catch (UnknownTaskException ignore) {
+            return defaultValue
+        }
+    }
```
labkey-tchad, line 56: For `Distribution.getArtifactId`
```
    static Optional<TaskProvider> getOptionalTask(Project project, String taskName)
    {
        try {
            return Optional.of(project.tasks.named(taskName))
        }
        catch (UnknownTaskException ignore) {
            return Optional.empty()
        }
    }
```

src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
```diff
@@ -233,10 +232,12 @@ class Distribution implements Plugin<Project>
...
             return project.dist.artifactId
-        else if (project.tasks.findByName("distribution") != null)
+        else
         {
-            if (project.tasks.distribution instanceof ModuleDistribution)
-                return ((ModuleDistribution) project.tasks.distribution).getArtifactId()
+            TaskUtils.doIfTaskPresent(project, "distribution", (distTask) -> {
+                if (distTask.get() instanceof ModuleDistribution)
+                    return ((ModuleDistribution) distTask).getArtifactId()
+            })
```
labkey-tchad: I don't think this will get returned from `getArtifactId`
```suggestion
            return TaskUtils.getOptionalTask(project, "distribution")
                    .filter(task -> task.get() instanceof ModuleDistribution)
                    .map(task -> ((ModuleDistribution)task.get()).getArtifactId())
                    .orElse(project.name)
```

src/main/groovy/org/labkey/gradle/plugin/JavaModule.groovy
```diff
@@ -185,43 +186,28 @@ class JavaModule implements Plugin<Project>
...
-                            allJars = allJars + project.tasks.apiJar.outputs.files
-                        }
-                        if (project.tasks.findByName('jspJar') != null)
-                        {
-                            allJars = allJars + project.tasks.jspJar.outputs.files
-                        }
+                        for (String taskName : JAR_TASK_NAMES)
+                            TaskUtils.doIfTaskPresent(project, taskName, (jarTask) -> {
+                                allJars = allJars + jarTask.get().outputs.files
+                            })
```
labkey-tchad, line 209: I didn't realize groovy doesn't share Java's restriction on accessing non-final variables within lambdas. Handy.

labkey-tchad (changes requested):
For places that could work with any of `project.layout.buildDirectory.file("file")`, `BuildUtils.getBuildDirFile(project, "file")`, or a `String`; is there an order of preference to which one to use?

labkey-susanh:
> For places that could work with any of `project.layout.buildDirectory.file("file")`, `BuildUtils.getBuildDirFile(project, "file")`, or a `String`; is there an order of preference to which one to use?

I tend to prefer the provider version if I know that it will work, simply because it seems to be the way Gradle is heading, but I also don't think it matters much. It annoys me that it's not at all obvious in various places where the string is needed vs. the file vs. the provider. I'm pretty sure the ant builder interface requires strings most of the time, for example.

src/main/groovy/org/labkey/gradle/plugin/JavaModule.groovy
labkey-susanh: Yeah. Still wish I hadn't chosen Groovy as the language here, though.

src/main/groovy/org/labkey/gradle/plugin/Distribution.groovy
labkey-susanh: I must not have tested with the right distribution. This code actually seems more complex than just the try-catch block, but OK.

labkey-tchad (approved):
Looks good. I just pushed a change to remove `TaskUtils.getFromTaskIfPresent`. They aren't actually used and `TaskUtils.getOptionalTask` can accomplish the same thing.

PR #187, opened 2023-12-13
Allow `application.properties` to define `server.tomcat.basedir`

labkey-tchad:
#### Rationale
To get access to all of the embedded server logs on TeamCity, we need to set `server.tomcat.basedir` to a known location (it defaults to a temp directory). Injecting `pathToServer` into `application.properties` is currently locked behind the `useLocalBuild` property. That property only works when building and running from a dev enlistment, some of the properties it enables don't work when deploying a distribution into a TeamCity test environment.

#### Related Pull Requests
- https://github.com/LabKey/server/pull/478

#### Changes
- Don't require `useLocalBuild` to inject server path into `application.properties`
- Update `log4j2.xml` and `labkey.xml` to include recent server changes

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ dependencies {
 }
 
 group 'org.labkey.build'
-project.version = "1.43.2-embeddedAccessLog-SNAPSHOT"
+project.version = "1.45.0-SNAPSHOT"
 
 gradlePlugin {
     // TODO after transitioning to using these plugin ids, remove the properties files from resources/META-INF.gradle-plugins
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ dependencies {
 }
 
 group 'org.labkey.build'
-project.version = "1.44.0-SNAPSHOT"
+project.version = "1.43.2-embeddedAccessLog-SNAPSHOT"
 
 gradlePlugin {
     // TODO after transitioning to using these plugin ids, remove the properties files from resources/META-INF.gradle-plugins
--- a/distributionResources/log4j2.xml
+++ b/distributionResources/log4j2.xml
@@ -21,8 +21,22 @@
             </PatternLayout>
             <Policies>
                 <OnStartupTriggeringPolicy />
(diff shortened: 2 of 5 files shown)


build.gradle
```diff
@@ -42,7 +42,7 @@ dependencies {
 }
 
 group 'org.labkey.build'
-project.version = "1.44.0-SNAPSHOT"
+project.version = "1.43.2-embeddedAccessLog-SNAPSHOT"
```
labkey-susanh: When you merge, please update to 1.45.0-SNAPSHOT as these changes seem like they are worth of a minor release change.

distributionResources/log4j2.xml
```diff
@@ -21,8 +21,22 @@
...
                 <OnStartupTriggeringPolicy />
+                <SizeBasedTriggeringPolicy size="10 MB"/>
             </Policies>
-            <DefaultRolloverStrategy max="3"/>
+            <DefaultRolloverStrategy fileIndex="min" max="3">
+                <Delete basePath="${sys:labkey.log.home}">
+                    <ScriptCondition>
+                        <!-- Use custom Java code to retain the first log file from the current session if it's about
+                        to be rotated out. It may have root cause information about problems that would otherwise be lost -->
+                        <Script name="retainFirst" language="rhino"><![CDATA[
```
labkey-susanh, line 31: rhino, eh? Do we know it's available where it's needed?

labkey-susanh (approved):

build.gradle
labkey-tchad: Sure. Can do.
BTW, these changes aren't super urgent if you want to include your `ignoreApiDep` changes in the release as well.

distributionResources/log4j2.xml
labkey-tchad: I don't think the "rhino" reference will be a problem. Normal installers built by TeamCity include the file from the server repo which has included [this script reference](https://github.com/LabKey/server/pull/124) for a couple of years now. These files are only included in standalone distributions.

PR #189, opened 2023-12-26
Prevent improper caching of `jsp2Java` task

labkey-tchad:
#### Rationale
When setting up builds for our Tomcat 10 migration, I had to disable the build cache because the `jsp2Java` task because it would pull cached results from our normal develop builds.

#### Related Pull Requests
- N/A

#### Changes
- Include `jspCompileClasspath` in `jsp2Java` task inputs to prevent bad cache hits

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ on how to do that, including how to develop and test locally and the versioning 
 _Note: 1.28.0 and later require Gradle 7_
 
 ### 1.44.1
-*Released*: TBD
+*Released*: 28 December 2023
 (Earliest compatible LabKey version: 23.3)
 * Prevent improper caching of `jsp2Java` task
 
--- a/README.md
+++ b/README.md
@@ -15,6 +15,11 @@ on how to do that, including how to develop and test locally and the versioning 
 
 _Note: 1.28.0 and later require Gradle 7_
 
+### 1.44.1
+*Released*: TBD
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -15,6 +15,11 @@ on how to do that, including how to develop and test locally and the versioning
 
 _Note: 1.28.0 and later require Gradle 7_
 
+### 1.44.1
+*Released*: TBD
```
labkey-susanh: If you want this published today, please update the release date before merging.

src/main/groovy/org/labkey/gradle/plugin/Jsp.groovy
```diff
@@ -173,6 +173,10 @@ class Jsp implements Plugin<Project>
                task.group = GroupNames.JSP
                task.description = "compile jsp files into Java classes"
 
+               if (project.hasProperty('apacheTomcatVersion'))
+                   task.inputs.property 'apacheTomcatVersion', project.property('apacheTomcatVersion')
+               else
+                   task.outputs.cacheIf { false } // Just don't cache if we can't be sure of the Jasper version
```
labkey-susanh, line 179: I'm not sure when this else case would happen, or if we'd actually be able to do much without a tomcat version, but not caching seems reasonable in that case.

labkey-susanh (approved):

PR #193, opened 2024-01-11
Update coordinates for gwt dependencies to `org.gwtproject` 

labkey-susanh:
#### Rationale
Seems that gwt artifacts have migrated from the group `com.google.gwt` to `org.gwtproject`. The latest version, which we need for our Tomcat10 conversion is available only at the new coordinates.

#### Changes
- Update dependency declarations for Gwt plugin

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/Gwt.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/Gwt.groovy
@@ -66,9 +66,16 @@ class Gwt implements Plugin<Project>
     {
         project.dependencies {
             gwtImplementation "org.gwtproject:gwt-user:${project.gwtVersion}",
-                    "org.gwtproject:gwt-dev:${project.gwtVersion}",
-                    "javax.validation:validation-api:${project.validationApiVersion}"
+                    "org.gwtproject:gwt-dev:${project.gwtVersion}"
         }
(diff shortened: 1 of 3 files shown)


src/main/groovy/org/labkey/gradle/plugin/Gwt.groovy
```diff
@@ -65,8 +65,8 @@ class Gwt implements Plugin<Project>
     private void addDependencies()
     {
         project.dependencies {
-            gwtImplementation "com.google.gwt:gwt-user:${project.gwtVersion}",
-                    "com.google.gwt:gwt-dev:${project.gwtVersion}",
+            gwtImplementation "org.gwtproject:gwt-user:${project.gwtVersion}",
+                    "org.gwtproject:gwt-dev:${project.gwtVersion}",
                     "javax.validation:validation-api:${project.validationApiVersion}"
```
labkey-adam: Have you confirmed that the validation coordinates are still correct?

labkey-adam (approved):

src/main/groovy/org/labkey/gradle/plugin/Gwt.groovy
labkey-susanh: Updated to choose the coordinates based on available properties

PR #195, opened 2024-01-29
Allow `useEmbeddedTomcat` and `useLocalBuild` to be turned off

labkey-tchad:
#### Rationale
In order to make embedded tomcat the default, we need a way to turn it off for testing on TeamCity.

#### Related Pull Requests
- N/A

#### Changes
- Allow `useEmbeddedTomcat` and `useLocalBuild` to be turned off with a "false" value

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ on how to do that, including how to develop and test locally and the versioning 
 _Note: 1.28.0 and later require Gradle 7_
 
 ### 2.2.2
-*Released*: TBD
+*Released*: 31 January 2024
 (Earliest compatible LabKey version: 24.2)
 * Allow `useEmbeddedTomcat` and `useLocalBuild` to be turned off
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,12 @@ on how to do that, including how to develop and test locally and the versioning 
 
 _Note: 1.28.0 and later require Gradle 7_
 
-### 2.1.1
+### 2.2.2
+*Released*: TBD
+(Earliest compatible LabKey version: 24.2)
+* Allow `useEmbeddedTomcat` and `useLocalBuild` to be turned off
+
+### 2.2.1
 *Released*: 22 January 2024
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -15,7 +15,12 @@ on how to do that, including how to develop and test locally and the versioning
 
 _Note: 1.28.0 and later require Gradle 7_
 
-### 2.1.1
+### 2.2.2
+*Released*: TBD
+(Earliest compatible LabKey version: 24.2)
+* Allow `useEmbeddedTomcat` and `useLocalBuild` to be turned off
+
+### 2.2.1
```
labkey-susanh, line 23: Below this, can you add something like
```
### versions accidentally skipped
```
so we don't wonder what happened.

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -882,7 +882,7 @@ class BuildUtils
 
     static boolean useEmbeddedTomcat(Project project)
     {
-        project.hasProperty(USE_EMBEDDED_TOMCAT)
+        project.hasProperty(USE_EMBEDDED_TOMCAT) && "false" != project.property(USE_EMBEDDED_TOMCAT)
```
labkey-susanh: You'll want to update the raw usages of `project.hasProperty("useEmbeddedTomcat")` in the root `settings.gradle` file and the bootstrap `build.gradle` file  to use this method as well.

labkey-susanh (approved):

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-tchad: `bootstrap/build.gradle` already uses this and `settings.gradle` doesn't have a `Project` to pass to this. I've created an overload that takes a `Settings` and performs a similar calculation if you'd like to re-review.

PR #198, opened 2024-02-22
Add property to force modules to be built from source

labkey-tchad:
#### Rationale
Sometime we remove modules and don't realize that they were still being referenced somewhere until we bump `labkeyVersion`. We are going to move a large number of modules soon so we should have some mechanism to prevent builds on TeamCity from pulling pre-built modules when we don't expect them to.

#### Related Pull Requests
- 

#### Changes
- Add property to force modules to be built from source

### DIFF
--- a/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
+++ b/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
@@ -266,8 +266,8 @@ class BuildUtils
             if (isSvnModule(project))
                 reasons.add("svn module without ${property} property set to true")
         }
-        else if (!Boolean.valueOf((String) value))
-            reasons.add("${property} property is false")
+        else if (BuildFromSource.fromProperty(value) == BuildFromSource._FALSE)
+            reasons.add("${property} property is not 'true'")
 
         return reasons
     }
@@ -703,6 +703,9 @@ class BuildUtils
         {
             if (depProject != null)
                 parentProject.logger.debug("Found project ${depProjectPath}; building ${depProjectPath} from source")
+            else
+                parentProject.logger.debug("Did not find project ${depProjectPath}; forcing project dependency as requested by '-P${BUILD_FROM_SOURCE_PROP}'")
+
             if (depProjectConfig != null)
                 parentProject.dependencies.add(parentProjectConfig, parentProject.dependencies.project(path: depProjectPath, configuration: depProjectConfig, transitive: isTransitive), closure)
             else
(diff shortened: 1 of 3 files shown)


labkey-susanh:
Do we need a new property? Properties passed through on the command line override properties defined within the modules themselves, and we are generally using `-PbuildFromSource=%gradle.buildFromSource%` in the TeamCity builds.

labkey-tchad:
> Do we need a new property? Properties passed through on the command line override properties defined within the modules themselves, and we are generally using `-PbuildFromSource=%gradle.buildFromSource%` in the TeamCity builds.

We are using `-PbuildFromSource=true` on TeamCity but it doesn't accomplish what I'm trying to do with this PR. If a module's project doesn't exist, `BuildUtils.addLabKeyDependency` doesn't even check that property; it just falls back to pulling modules from Artifactory.
Is `buildFromSource` not working as intended? I didn't want to change how it works because its used in several other places.

labkey-susanh:
> > Do we need a new property? Properties passed through on the command line override properties defined within the modules themselves, and we are generally using `-PbuildFromSource=%gradle.buildFromSource%` in the TeamCity builds.
> 
> We are using `-PbuildFromSource=true` on TeamCity but it doesn't accomplish what I'm trying to do with this PR. If a module's project doesn't exist, `BuildUtils.addLabKeyDependency` doesn't even check that property; it just falls back to pulling modules from Artifactory. Is `buildFromSource` not working as intended? I didn't want to change how it works because its used in several other places.

That is as intended, yes. One more thought: would allowing for `buildFromSource=force` accomplish what you're looking for here?

labkey-tchad:
> That is as intended, yes. One more thought: would allowing for `buildFromSource=force` accomplish what you're looking for here?

Ok. I've updated it to recognize `buildFromSource=force`

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -964,4 +973,22 @@ class BuildUtils
...
+        static BuildFromSource fromProperty(Project project, String propertyName) {
+            String propertyValue = project.hasProperty(propertyName) ? project.property(propertyName) : null
+            fromProperty(propertyValue)
+        }
+
+        static BuildFromSource fromProperty(String propertyValue) {
+            if ('force' == propertyValue)
+                return FORCE
+            else
+                return Boolean.valueOf(propertyValue) ? _TRUE : _FALSE
```
labkey-susanh, line 991: This is changing the behavior of the current property. Currently, the property has to be explicitly set to false to turn off building from source. Otherwise, if the property is there or set to true we will build from source.

labkey-tchad: Maybe I'm missing something. The only place that currently checks the value of `buildFromSource` is in `BuildUtils.whyNotBuildFromSource`. It also uses `Boolean.valueOf()` which only returns `true` for `value.equalsIgnoreCase("true")`:
```
else if (!Boolean.valueOf((String) value))
    reasons.add("${property} property is false")
```
It has been updated to use this enum and should behave the same way:
```
else if (BuildFromSource.fromProperty(value) == BuildFromSource._FALSE)
    reasons.add("${property} property is not 'true'")
```

labkey-susanh: Ah you are right. Too much Javascript in my brain.

labkey-susanh (approved):

PR #201, opened 2024-03-08
Issue 49824: Improve embedded Tomcat auto-redeploy using a trigger file

labkey-susanh:
#### Rationale
[Issue 49824](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49824) Currently, we have configured the restart of embedded tomcat by having SpringBoot watch the `build/deploy/modules` and `build/deploy/embedded` directories. Since we write a lot of things to these directories during a deployment, when using a local build, usually the restart from SpringBoot happens before all the copying is complete, resulting in a failed start and the need to start manually.  This PR adds a final step in various tasks that are meant to trigger a restart , which will modify a file that we will configure SpringBoot to look for as a signal to restart (see [SpringBoot docs](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#using.devtools.restart.triggerfile)).

#### Related Pull Requests
- https://github.com/LabKey/server/pull/759

#### Changes
- Add a `BuildUtils.updateRestartTriggerFile` method and tie it into various restart-triggering tasks

### DIFF
--- a/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
+++ b/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
@@ -914,7 +914,7 @@ class BuildUtils
      * We use build/deploy/modules because when using a local build it's added in the application.properties file as a
      * spring.devtools.restart.additional-paths
      *
-     * @param project
+     * @param project - for use in getting the rootProject's build directory
      */
     static void updateRestartTriggerFile(Project project)
     {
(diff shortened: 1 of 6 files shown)


src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -904,6 +906,31 @@ class BuildUtils
         o.hasProperty(USE_EMBEDDED_TOMCAT) && o[USE_EMBEDDED_TOMCAT] != "false"
     }
 
+    /**
+     * Writes a file in the build/deploy/modules directory that can be used as a trigger file for restarting
+     * SpringBoot. Without this, restarts may happen before the full application deployment is done, resulting
+     * if a failed start. See
```
labkey-jeckels:
```suggestion
     * in a failed start. See
```

labkey-jeckels (approved):

PR #204, opened 2024-03-12
Don't create reload trigger file if output directory doesn't exist

labkey-susanh:
#### Rationale
Since the reload trigger file is not designated as an output file of any of the tasks that use this utility method, the directory does not automatically get created. If there is no `build/deploy/modules` directory, there is no restart to be triggered, so we'll just exit from the method that wants to update that file.

#### Related Pull Requests
- 

#### Changes
- Update `BuildUtils.updateRestartTriggerFile` to check for output directory before trying to write the file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ _Note: 1.28.0 and later require Gradle 7_
 ### TBD
 *Released*: TBD
 (Earliest compatible LabKey version: 24.2)
-* Create output directory for reloadTrigger file if it doesn't exist
+* Account for missing directory for reloadTrigger fle
 
 ### 2.6.0
 *Released*: 11 March 2024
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -15,6 +15,11 @@ on how to do that, including how to develop and test locally and the versioning
 
 _Note: 1.28.0 and later require Gradle 7_
 
+### TBD
+*Released*: TBD
+(Earliest compatible LabKey version: 24.2)
+* Account for missing directory for reloadTrigger fle
```
labkey-danield, line 21:
```suggestion
* Account for missing directory for reloadTrigger file
```

labkey-danield: I'm assuming that is a typo.

labkey-danield (approved):

labkey-susanh:
Branch got weird on me.

PR #206, opened 2024-04-05
Hack to fix BuildUtils.getModuleProjectPath to account for limsModules repository

labkey-susanh:
#### Rationale
The `BuildUtils.getModuleProjectPath` assumed all modules were at the top level, but that is not true. This fix is a hack that does enough for our 24.3 release purposes. We'll want to change the dependency exclusion declaration to move away from using the project path and use either the group and artifact designation instead (e.g., "org.labkey.module:biologics") or probably just the module name, but this change is sufficient for current purposes.

#### Related Pull Requests
- 

#### Changes
- Special case for limsModules in `BuildUtils.getModuleProjectPath`

### DIFF
--- a/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
+++ b/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
@@ -551,6 +551,8 @@ class BuildUtils
             'tomcat7-websocket'
     ]
 
+    private static List LIMS_MODULES = ["biologics", "inventory", "labbook", "puppeteer", "recipe", "sampleManagement"]
+
     static String getGitUrl(Project project)
     {
         def grgit = Grgit.open(currentDir: project.projectDir)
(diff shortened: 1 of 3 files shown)


src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -788,6 +790,8 @@ class BuildUtils
 
     static String getModuleProjectPath(ModuleDependency dependency)
     {
+        if (LIMS_MODULES.contains(dependency.getName()))
+            return ":server:modules:limsModules:" + dependency.getName()
```
labkey-jeckels, line 794: Are there any similar references to `premiumModules` that we should hack too?

labkey-jeckels (approved):

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-susanh: I found no other references.

PR #209, opened 2024-05-13
Add and update flags for embedded tomcat

labkey-tchad:
#### Rationale
Adding override flags for distribution types will allow us to create non-standard distributions more easily.
Adding the required `--add-opens` parameters to the `startTomcat` task for embedded tomcat will make it function correctly by default without having to use the `extraCatalinaOpts` parameter.

#### Related Pull Requests
- 

#### Changes
* Include required reflection parameters to embedded `StartTomcat`
* Build embedded distribution archives regardless of `useEmbeddedTomcat` flag
* Add flags to override distributions' specified archive type (`forceStandaloneDist` & `forceEmbeddedDist`)

### DIFF
--- a/src/main/groovy/org/labkey/gradle/task/StartTomcat.groovy
+++ b/src/main/groovy/org/labkey/gradle/task/StartTomcat.groovy
@@ -174,7 +174,7 @@ class StartTomcat extends DefaultTask
         }
 
         if (project.hasProperty("extraCatalinaOpts"))
-            optsList.addAll(((String) project.property("extraCatalinaOpts")).split(" "))
+            optsList.addAll(((String) project.property("extraCatalinaOpts")).split("\\s+"))
 
         return optsList.stream()
                 .filter({String opt -> return !StringUtils.isEmpty(opt)})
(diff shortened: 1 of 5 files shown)


src/main/groovy/org/labkey/gradle/task/StartTomcat.groovy
```diff
@@ -171,4 +181,14 @@ class StartTomcat extends DefaultTask
                 .collect(Collectors.toList())
 
     }
+
+    private static List<String> getEmbeddedReflectionOpts(Project project)
+    {
+        if (project.hasProperty(EMBEDDED_REFLECTION_PARAM)) {
+            return ((String) project.property(EMBEDDED_REFLECTION_PARAM)).trim().split(" +")
```
labkey-susanh: The split on " +" seems unconventional. Why the "+"?

labkey-tchad: Regex to catch multiple spaces. Probably not strictly necessary but should be safe.

labkey-susanh: I  always see spaces as `\s` in regular expressions, so am surprised it works.

labkey-susanh (approved):

PR #212, opened 2024-07-12
Remove support for AntBuild and resources jsps and make some headway on compatibility with configuration cache.

labkey-susanh:
#### Rationale
When we converted to Gradle from Ant, we took advantage of Gradle's ability to incorporate Ant builds as well to ease the transition to the new build technology. That was done some many years ago now, so it is time to remove support for Ant builds explicitly.

We had, at some point, some modules that had `jsp` files in their `resources` directories. This does not appear to be the case anymore, so we'll get rid of the task that was looking for these files and just assume all `jsp` files are under the `src` directory.

The Gradle folks have been hard at work on their [configuration cache](https://docs.gradle.org/8.8/userguide/configuration_cache.html) feature, and plan to make this a feature that is on by default when Gradle 9 is released and not optional at some point after that. The configuration phase of the build takes significant time, so using this cache can save some real time. Using it does require changing some patterns we've used in many of our plugins and tasks, mostly because we reference the `project` object within task execution, which makes the task configuration not cacheable.  With this PR, I've converted some of the tasks that previously were not compatible and made them compatible and I've also made some annotations using the `notCompatibleWithConfigurationCache` method to indicate some of the additional work required. (There's more work to do than is indicated by these annotations, so don't be fooled.)

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5669
- https://github.com/LabKey/server/pull/848

#### Changes
- Gradle 8.9
- Remove `AntBuild` plugin
- Updates for compatibility with configuration cache
  - `SchemaCompile`
  - `WriteDependenciesFile`
  - Most of the Jsp plugin, except where not supported from Gradle or not in usual build chain
  - `CheckForVersionConflicts` 
  - `ModuleXmlFile` task introduced for ad-hoc `moduleXml` task
- Remove support for `.jsp` files in module `resources` directories
- Remove support for `creditable` configuration for generating entries in `jars.txt` file
- Remove `Standalone` plugin for building stand-alone distributions


labkey-adam (approved):

src/main/groovy/org/labkey/gradle/task/JspCompile2Java.groovy
```diff
@@ -18,29 +18,38 @@ package org.labkey.gradle.task
...
-        return BuildUtils.getBuildDirFile(project, CLASSES_DIR)
+    final abstract DirectoryProperty classesDirectory = project.objects.directoryProperty().convention(project.layout.buildDirectory.dir(CLASSES_DIR).get())
+
+    @InputFiles
+    @CompileClasspath
+    abstract ConfigurableFileCollection getCompileClasspath()
+
+    @Inject
+    JspCompile2Java(FileSystemOperations fs) {
+        fileSystemOperations = fs
```
labkey-tchad, line 49: Is this functionally different from the way the file system is injected in `SchemaCompile` or just a different way to accomplish the same thing?
```
@Inject abstract FileSystemOperations getFs()
```

labkey-tchad (approved):
Removing all project references from tasks is going to a "fun" project 😅

src/main/groovy/org/labkey/gradle/task/JspCompile2Java.groovy
labkey-susanh: It is, as far as I know, just a different way to accomplish the same thing. My first attempts to do the injection via the getter didn't seem to work, so I went with the constructor route.  I might be able to switch this, but decided to leave this in to serve as an example for my future self.

PR #213, opened 2024-07-18
Add missing input for `CopyJsp` and missing `onlyIf` config for `WriteDependenciesFile`

labkey-susanh:
#### Rationale
The previous PR failed to configure the `CopyJsp` task properly so changes to individual `Jsp` files were not being picked up (not sure this change to requiring the use of FileSystem injection is altogether a good thing...). 

The previous PR also failed to prevent the `WriteDependenciesFile` task from running when a module, such as a file-based module, has no dependencies to write about.

#### Related Pull Requests
- #212 

#### Changes
- Add `InputDirectory` for `CopyJsp` task
- Add `onlyIf` config for `WriteDependenciesFile` task

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/ModuleResources.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/ModuleResources.groovy
@@ -42,9 +42,6 @@ class ModuleResources
                             task.getArtifactIds().set(artifacts.map(new WriteDependenciesFile.IdExtractor()))
                     }
                     task.externalDependencies.set(project.extensions.findByType(ModuleExtension.class).getExternalDependencies())
-                    task.onlyIf {
-                        return !task.externalDependencies.get().isEmpty()
-                    }
                 } catch (UnknownDomainObjectException ignore) {
 
                 }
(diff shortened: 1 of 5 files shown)


src/main/groovy/org/labkey/gradle/plugin/ModuleResources.groovy
```diff
@@ -41,10 +41,13 @@ class ModuleResources
                             Provider<Set<ResolvedArtifactResult>> artifacts = config.getIncoming().getArtifacts().getResolvedArtifacts();
                             task.getArtifactIds().set(artifacts.map(new WriteDependenciesFile.IdExtractor()))
                     }
+                    task.externalDependencies.set(project.extensions.findByType(ModuleExtension.class).getExternalDependencies())
+                    task.onlyIf {
+                        return !task.externalDependencies.get().isEmpty()
+                    }
                 } catch (UnknownDomainObjectException ignore) {
 
                 }
```
labkey-tchad: Would an `UnknownDomainObjectException` cause this task to run when it shouldn't since `task.onlyIf {...}` won't be reached?
Should the `onlyIf` just be baked into `WriteDependenciesFile`, similar to `InstallRPackage`?

labkey-tchad (approved):
One question

src/main/groovy/org/labkey/gradle/plugin/ModuleResources.groovy
labkey-susanh: Technically, that's two questions. 😄, but they are both good questions, so I'll let it slide. 

Effectively, it won't happen anymore that the exception is thrown since the configuration is created more broadly, but still a good idea to check. I had assumed Gradle wouldn't like the `onlyIf` inside the class, but that's cleaner, so I'll try it.

labkey-tchad: There is no escape from off-by-one errors 😰

PR #215, opened 2024-08-22
Update distributions task to build embedded Tomcat tar.gz archives only

labkey-adam:
#### Rationale
We only build embedded Tomcat distributions now. Clear out unused properties and options.

#### Related Pull Requests
- https://github.com/LabKey/distributions/pull/473

### DIFF
--- a/src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
+++ b/src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
@@ -22,6 +22,7 @@ import org.gradle.api.Project
 import org.gradle.api.file.CopySpec
 import org.gradle.api.file.DuplicatesStrategy
 import org.gradle.api.file.FileTree
+import org.gradle.api.provider.Property
 import org.gradle.api.tasks.*
 import org.labkey.gradle.plugin.ApplyLicenses
 import org.labkey.gradle.plugin.extension.DistributionExtension
@@ -42,7 +43,6 @@ class ModuleDistribution extends DefaultTask
     @Optional @Input
     String versionPrefix = null
     @Optional @Input
-    @Input
     final abstract Property<String> subDirName = project.objects.property(String).convention(project.name)
     @Optional @Input
     String archivePrefix = "LabKey"
@@ -103,7 +103,7 @@ class ModuleDistribution extends DefaultTask
 
         createDistributionFiles()
         gatherModules()
-        packageRedistributables()
+        embeddedTomcatTarArchive()
     }
 
     @OutputDirectory
@@ -153,11 +153,6 @@ class ModuleDistribution extends DefaultTask
         }
     }
 
-    private void packageRedistributables()
-    {
-        embeddedTomcatTarArchive()
-    }
-
     @Input
     String getArtifactId()
     {
@@ -186,11 +181,6 @@ class ModuleDistribution extends DefaultTask
     private String getEmbeddedTarArchivePath()
     {
         return "${getDistributionDir()}/${getArchiveName()}${DistributionExtension.EMBEDDED_SUFFIX}.${DistributionExtension.TAR_ARCHIVE_EXTENSION}"
-    }
-
-    private String getWarArchivePath()
-    {
-        return "${getDistributionDir()}/${getArchiveName()}.war"
     }
 
     private File getWindowsUtilDir()
--- a/src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
+++ b/src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
@@ -28,30 +28,25 @@ import org.labkey.gradle.plugin.extension.DistributionExtension
 import org.labkey.gradle.plugin.extension.LabKeyExtension
 import org.labkey.gradle.plugin.extension.StagingExtension
 import org.labkey.gradle.util.BuildUtils
-import org.labkey.gradle.util.PropertiesUtils
 
 import java.nio.file.Files
 
 class ModuleDistribution extends DefaultTask
 {
-    @Optional @Input
-    Boolean includeZipArchive = false
-    @Optional @Input
-    Boolean includeTarGZArchive = false
+    // TODO: Remove this unused property
     @Optional @Input
     String embeddedArchiveType = null
-    @Optional @Input
-    Boolean makeDistribution = true // set to false for just an archive of modules
+
     @Optional @Input
     String extraFileIdentifier = ""
     @Optional @Input
     String versionPrefix = null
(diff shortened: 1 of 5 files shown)


labkey-adam:
@labkey-susanh everything builds just fine with all our current `build.gradle` files, but I could use help with one thing: getting `subDirName` optional behavior to work right. If this property is removed from `build.gradle`, I see the following error:
```
FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':distributions:starter:distribution' (type 'ModuleDistribution').
  - In plugin 'org.labkey.build.base' type 'org.labkey.gradle.task.ModuleDistribution' property 'artifactId' doesn't have a configured value.

    Reason: This property isn't marked as optional and no value has been configured.

    Possible solutions:
      1. Assign a value to 'artifactId'.
      2. Mark property 'artifactId' as optional.
```

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -28,30 +28,25 @@ import org.labkey.gradle.plugin.extension.DistributionExtension
...
-    @Optional @Input
-    Boolean makeDistribution = true // set to false for just an archive of modules
+
     @Optional @Input
     String extraFileIdentifier = ""
     @Optional @Input
     String versionPrefix = null
     @Optional @Input
-    String subDirName
+    String subDirName = null
```
labkey-susanh:
```suggestion
    @Input
    final abstract Property<String> subDirName = project.objects.property(String).convention(project.name)
```

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -80,45 +75,21 @@ class ModuleDistribution extends DefaultTask
     File getDistributionDir()
     {
         if (distributionDir == null) {
-            var subDir = StringUtils.trimToNull(subDirName)
-            if (subDir == null)
-                subDir = project.name
-            distributionDir = project.file("${distExtension.dir}/${subDir}")
+            distributionDir = project.file("${distExtension.dir}/" + getSubDirName())
```
labkey-susanh:
```suggestion
            distributionDir = project.file("${distExtension.dir}/" + subDirName.get())
```

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -185,47 +154,13 @@ class ModuleDistribution extends DefaultTask
...
-    {
-        if (shouldBuildStandaloneTarGz())
-            tarArchives()
-        if (shouldBuildEmbeddedArchive(DistributionExtension.TAR_ARCHIVE_EXTENSION))
-            embeddedTomcatTarArchive()
-        if (includeZipArchive)
-            zipArchives()
-        if (shouldBuildEmbeddedArchive(DistributionExtension.ZIP_ARCHIVE_EXTENSION))
-            embeddedTomcatZipArchive()
+        embeddedTomcatTarArchive()
```
labkey-susanh: Seems that this method could just be inlined.

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -185,47 +154,13 @@ class ModuleDistribution extends DefaultTask
...
     }
 
     @Input
     String getArtifactId()
     {
-        return subDirName
+        return getSubDirName()
```
labkey-susanh:
```suggestion
        return subDirName.get()
```

labkey-susanh (commented):
I think using a Property with a convention to provide the default value may get you what you need for the subDirName.

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -247,26 +182,11 @@ class ModuleDistribution extends DefaultTask
...
-    }
-
-    private String getEmbeddedZipArchivePath()
-    {
-        return "${getDistributionDir()}/${getArchiveName()}${DistributionExtension.EMBEDDED_SUFFIX}.${DistributionExtension.ZIP_ARCHIVE_EXTENSION}"
-    }
-
     private String getWarArchivePath()
     {
         return "${getDistributionDir()}/${getArchiveName()}.war"
```
labkey-tchad: Maybe remove `getWarArchivePath`. I think we abandoned that a long time ago.

labkey-tchad (approved):

labkey-adam:
I think everything is fixed up now. Local testing worked with `subDirName` omitted. @labkey-susanh feel free to re-review if you want.

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -22,36 +22,32 @@ import org.gradle.api.Project
...
     @Optional @Input
     String embeddedArchiveType = null
-    @Optional @Input
-    Boolean makeDistribution = true // set to false for just an archive of modules
+
     @Optional @Input
     String extraFileIdentifier = ""
     @Optional @Input
     String versionPrefix = null
     @Optional @Input
```
labkey-susanh, line 45: I think the `@Optional` can/should be removed here. We need some kind of name.

labkey-susanh (approved):

src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
labkey-adam: Interesting... this was marked `@Optional` before my changes. What about keeping it optional and defaulting to `project.name` a la `subDirName`?

labkey-susanh: I thought I was pointing to the `@Optional` for `subDirName` with this comment.   Did I misread?

labkey-adam: Sorry, I was unclear which `@Optional` was being referenced. As for `subDirName`, the code was clearly attempting to make it optional; it was annotated as such and defaulting to `project.name.` Only problem was `getArtifactId()` hadn't gotten the memo. It now correctly defaults to `project.name`, which was my goal and what we discussed over email. But maybe I'm misunderstanding your suggestion or what `@Optional` means...

labkey-susanh: Oh, my bad. I think we do want to leave it so Gradle/Groovy knows the task doesn't require a value to be provided.

PR #216, opened 2024-09-04
Remove support for useEmbeddedTomcat, tomcat-lib, and log4j2.xml

labkey-adam:
#### Rationale
No need to handle standalone builds


src/main/groovy/org/labkey/gradle/plugin/ServerDeploy.groovy
```diff
@@ -432,92 +375,6 @@ class ServerDeploy implements Plugin<Project>
...
-    }
-
-    private void deployTomcatJars(Task task) {
-        Project project = task.project
-        if (BuildUtils.useEmbeddedTomcat(project)) {
-            task.logger.info("Not deploying tomcat jars for embedded deployment")
-            return
-        }
-
-        JDBC_JARS.each{String name -> new File("${project.tomcat.catalinaHome}/lib/${name}").delete()}
```
labkey-susanh, line 507: Probably can get rid of this `JDBC_JARS` constant too.

src/main/groovy/org/labkey/gradle/plugin/ServerDeploy.groovy
```diff
@@ -558,28 +415,6 @@ class ServerDeploy implements Plugin<Project>
...
-        }
-
-        project.configurations.tomcatJars.files.each {File jarFile ->
-            File libFile = new File("${project.tomcat.catalinaHome}/lib/${jarFile.getName()}")
-            if (libFile.exists())
-                project.delete libFile.getAbsolutePath()
-        }
-
-        // Get rid of (un-versioned) jars that were deployed
-        TOMCAT_LIB_UNVERSIONED_JARS.each{String name ->
```
labkey-susanh, line 577: Probably can get rid of this `TOMCAT_LIB_UNVERSIONED_JARS` constant.

src/main/groovy/org/labkey/gradle/task/RunUiTest.groovy
```diff
@@ -88,11 +88,10 @@ abstract class RunUiTest extends Test
         systemProperty "labkey.root", project.rootProject.projectDir
         systemProperty "project.root", project.rootProject.projectDir
         systemProperty "user.home", System.getProperty('user.home')
-        // A handfull of tests require tomcat.home to be defined when running within IntelliJ
+        // A handful of tests require tomcat.home to be defined when running within IntelliJ
         systemProperty "tomcat.home", project.tomcat.catalinaHome
         systemProperty "test.credentials.file", "${project.projectDir}/test.credentials.json"
-        if (BuildUtils.useEmbeddedTomcat(project))
-            systemProperty BuildUtils.USE_EMBEDDED_TOMCAT, ''
+        systemProperty BuildUtils.USE_EMBEDDED_TOMCAT, ''
```
labkey-susanh: I wonder if this is still needed. Isn't embedded the default now?

src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
```diff
@@ -108,10 +108,6 @@ class TeamCity extends Tomcat
                 task.group = GroupNames.TEST_SERVER
                 task.description = "Removes log files from Tomcat and TeamCity"
                 task.dependsOn project.tasks.cleanLogs
-                if (!BuildUtils.useEmbeddedTomcat(project)) {
-                    // Not valid for embedded Tomcat
-                    task.dependsOn project.tasks.cleanTemp
```
labkey-susanh, line 113: Can this task definition (in the `Tomcat` plugin) be removed if not valid for embedded?  That would also cascade to removing at least the `validateCatalinaHome` method from the `TomcatExtension`.

src/main/groovy/org/labkey/gradle/plugin/ServerDeploy.groovy
```diff
@@ -364,32 +333,6 @@ class ServerDeploy implements Plugin<Project>
         }
         project.tasks.named('deployApp').configure {mustRunAfter(project.tasks.cleanBuild)}
```
labkey-susanh, line 322: I think we can also remove, somewhere close to here, the lines that check for catalinaHome and hasPlugin(Tomcat) since CATALINA_HOME isn't relevant for embedded and we aren't supporting the standalone builds anymore:
```
if (project.plugins.hasPlugin(Tomcat)) {
            if (project.tomcat.hasCatalinaHome()) {
                project.tasks.named("cleanBuild").configure {
                    it.dependsOn(project.tasks.stopTomcat)
                }
                project.tasks.named("cleanDeploy").configure {
                    it.dependsOn(project.tasks.stopTomcat)
                }
            }
        }
```
I think this also cascades to removing the `hasCatalinaHome` method from `TomcatExtension`.

labkey-susanh (commented):
I think there are some methods and/or properties that

src/main/groovy/org/labkey/gradle/task/DeployApp.groovy
```diff
@@ -47,25 +44,12 @@ class DeployApp extends DeployAppBase
...
         deployPlatformBinaries(deployBinDir)
     }
 
-
-    private void deployWebappDir()
-    {
-        ant.copy(
-                todir: deployWebappDir,
-                preserveLastModified: true
-        )
```
labkey-tchad, line 63: We still need a `labkeywebapp` directory in our deployments. Will this directory get created in some other way?

src/main/groovy/org/labkey/gradle/task/RunUiTest.groovy
```diff
@@ -88,11 +88,9 @@ abstract class RunUiTest extends Test
         systemProperty "labkey.root", project.rootProject.projectDir
         systemProperty "project.root", project.rootProject.projectDir
         systemProperty "user.home", System.getProperty('user.home')
-        // A handfull of tests require tomcat.home to be defined when running within IntelliJ
+        // A handful of tests require tomcat.home to be defined when running within IntelliJ
         systemProperty "tomcat.home", project.tomcat.catalinaHome
```
labkey-tchad: In my full enlistment, there is only one reference to `tomcat.home`, in a method that is only called twice (`TestProperties.getTomcatHome`). Your testAutomation PR removes one and should remove the other as well (I left feedback in that PR)
```suggestion
```

src/main/groovy/org/labkey/gradle/plugin/TeamCity.groovy
labkey-adam: Removed, which led to removing what seems like the primary use of `TomcatExtension` (in `Tomcat.groovy`). One reference remains, in `RunUiTest.groovy`. Unclear if we can remove that as well (and the extension).

labkey-tchad (changes requested):
We can probably get rid of the entire `distributionResources` directory and any tasks that publish it. The `distributionResources/embedded` is just an incomplete duplicate of `server/configs/webapps/embedded`. The files here are only used if the corresponding files from the server repo are missing -- which should never happen. Everything else in `distributionResources` is for standalone tomcat.

src/main/groovy/org/labkey/gradle/task/DeployApp.groovy
labkey-adam: `/deploy/labkeyWebapp` is still getting created locally. I think this task was just copying the `/staging/webapp` files, of which there are now none.

src/main/groovy/org/labkey/gradle/plugin/ServerDeploy.groovy
```diff
@@ -420,102 +359,6 @@ class ServerDeploy implements Plugin<Project>
...
         project.tasks.named('deployApp').configure {dependsOn(project.tasks.named("checkModuleTasks"))}
         project.tasks.named('checkModuleTasks').configure {mustRunAfter(project.tasks.stageApp)} // do this so the message appears at the bottom of the output
-        if (project.plugins.hasPlugin(Tomcat)) {
-            if (project.tomcat.hasCatalinaHome()) {
-                project.tasks.named("cleanBuild").configure {
-                    it.dependsOn(project.tasks.stopTomcat)
-                }
-                project.tasks.named("cleanDeploy").configure {
-                    it.dependsOn(project.tasks.stopTomcat)
-                }
```
labkey-susanh, line 430: Have you tried `cleanBuild` and/or `cleanDeploy` without this dependency on `stopTomcat`. I'm guessing we still want the task dependencies, just don't need the if statements around them.

src/main/groovy/org/labkey/gradle/plugin/extension/TomcatExtension.groovy
```diff
@@ -49,36 +48,6 @@ class TomcatExtension
         return tomcatConfDir
     }
 
-    void validateCatalinaHome()
```
labkey-susanh, line 52: @labkey-tchad Are the remaining references to the properties in `TomcatExtension` that are designated as used by TeamCity still relevant?

labkey-susanh (approved):
Might still be able to clean up that TomcatExtension, but I'm going to leave that for Trey to answer (and we can do it later if we find it's not used).

src/main/groovy/org/labkey/gradle/plugin/extension/TomcatExtension.groovy
labkey-tchad: They are not. I have a PR into this branch to remove them all: #217

src/main/groovy/org/labkey/gradle/plugin/ServerDeploy.groovy
labkey-adam: I restored the two `dependsOn(stopTomcat)` calls (now unconditional)

labkey-tchad (approved):

PR #218, opened 2024-09-16
Strip "_dist" suffix from project.name if present

labkey-adam:
#### Rationale
Produce more useful `subDirName` and `extraFileIdentifier` property default values for `project.names` such as `nci_dist`, `cds_dist`, etc. This will eventually help with https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=47603

### DIFF
--- a/src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
+++ b/src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
@@ -21,7 +21,6 @@ import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.file.CopySpec
 import org.gradle.api.file.DuplicatesStrategy
-import org.gradle.api.provider.Property
 import org.gradle.api.tasks.*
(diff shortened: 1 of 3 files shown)


src/main/groovy/org/labkey/gradle/task/ModuleDistribution.groovy
```diff
@@ -150,19 +149,30 @@ class ModuleDistribution extends DefaultTask
...
         }
         return archiveName
     }
 
+    private String getSubDir()
+    {
+        return subDirName == null ? getProjectName() : subDirName;
+    }
+
+    private String getProjectName()
```
labkey-susanh: I would rather not call this `getProjectName` since that has a particular meaning for Gradle. Perhaps `getDistName` or something like that

labkey-adam: Changed to `getDefaultName()`

labkey-susanh (approved):

PR #219, opened 2024-09-21
Eliminate "-embedded" suffix from distribution filenames

labkey-adam:
#### Rationale
- All distributions are embedded, so no need to include this term in every filename. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=47603
- Infer appropriate "default database" value based on JDBC driver name and remove support for `databaseMaster` property. 
- Add support for `extraProperties` map, which gets written to `WEB-INF/classes/distribution.properties` along with standard distribution build properties
- Assign `group` inside `ModuleDistribution` task so build files no longer need to
- Add `startLabKey` and `stopLabKey` tasks that will eventually replace `startTomcat` and `stopTomcat`
- Eliminate some old zip and embedded references


src/main/groovy/org/labkey/gradle/task/DeployEmbeddedDistribution.groovy
```diff
@@ -19,21 +18,19 @@ class DeployEmbeddedDistribution extends DeployAppBase {
...
     {
         deployExecutableJar()
-        deployNlpEngine(deployBinDir)
         deployPlatformBinaries(deployBinDir)
     }
 
     private void deployExecutableJar() {
-        File distributionFile = DistributionExtension.getDistributionFile(project, true)
-        boolean isTar = FilenameUtils.getExtension(distributionFile.getName()).equals("gz")
+        File distributionFile = DistributionExtension.getDistributionFile(project)
```
labkey-susanh, line 25: This task could probably be renamed to simply `DeployDistribution`. Or it could be left as is. Either seems fine.

labkey-susanh (approved):

src/main/groovy/org/labkey/gradle/task/DeployEmbeddedDistribution.groovy
labkey-adam: Done

FYI: I've removed a few other unnecessary "embedded" references, e.g., `getEmbeddedTomcatJarPath()` -> `getLabKeyServerJarPath()`, `getEmbeddedTarArchivePath()` -> `getTarArchivePath()`

PR #223, opened 2024-10-15
Need to restore the generic SNAPSHOT version

labkey-adam:
#### Rationale
I merged previous PR with the FB-specific version

#### Related Pull Requests
- https://github.com/LabKey/gradlePlugin/pull/222

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@
 The gradlePlugin jar is a jar file containing plugins, tasks, extensions and utilities used for building the [LabKey](https://www.labkey.org)
 Server application and its modules.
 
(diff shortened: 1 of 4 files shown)


labkey-susanh:
I think I'm missing something. I see all the changes from the previous PR but not an update to the version number, which is the only change I would expect given.

labkey-tchad (commented):
This doesn't seem to modify `build.gradle`. Maybe needs a merge from develop?

labkey-adam:
I was reusing the old branch... I'll make a new one so it's clearer

PR #225, opened 2024-11-04
Gradle 9 and configuration cache compatibility work

labkey-susanh:
#### Rationale
The `stageModules` task has been using a deprecated method for collecting the modules to be staged. This method is scheduled for removal in Gradle 9.0, so the main goal of this PR is to update that method to remove the use of said method.
I've also taken this opportunity to update more of the tasks to be compatible with the configuration cache and remove one task that is no longer very useful and not compatible with the cache.

#### Related Pull Requests
- https://github.com/LabKey/server/pull/921
- https://github.com/LabKey/DiscvrLabKeyModules/pull/322

#### Changes
- Remove `checkModuleTasks` tasks, added to get us through a transition from plugins being declared more centrally
- Update `stageModules` to remove use of deprecated `fileCollection` method and make compatible with configuration cache
- Remove `StagingExtension`
- Make `jsp2Java` compatible with configuration cache
- Update property in node plugin configuration for latest version
- Mark `copyExternalLibs` task as not compatible with configuration cache for now
- Update `ServerSideJS` task for configuration cache compatibility
- Update `DeployApp` and relatives for better configuration cache compatibility
- Update `PickDb` task for better configuration cache compatibility
- Upgrade to Gradle 8.10.2


src/main/groovy/org/labkey/gradle/task/RestartTriggerTask.groovy
```diff
@@ -0,0 +1,43 @@
...
+        OutputStreamWriter writer = null
+        try {
+            File triggerFile = new File(triggerFileDir.get().asFile, RESTART_FILE_NAME)
+            writer = new OutputStreamWriter(new FileOutputStream(triggerFile), StandardCharsets.UTF_8)
+            writer.write(SimpleDateFormat.getDateTimeInstance().format(new Date()))
+        }
+        finally
+        {
+            if (writer != null)
+                writer.close()
```
labkey-adam: Just curious: does Groovy have try-with-resources?

labkey-adam (approved):

src/main/groovy/org/labkey/gradle/task/RestartTriggerTask.groovy
labkey-susanh: Last time I tried, it did not, but I can try again.

labkey-adam: I don't really care, it just seems funny to see this pattern these days

labkey-susanh: Seems like it is now supported. (The lag in language support is certainly one of the reasons I regret having chosen Groovy in the first place. Someday maybe we can make the switch to Java (which Gradle itself did a long time ago).)

labkey-adam: I would certainly support that change! Could a Groovy --> Java migration happen incrementally? Or is it all or nothing?

labkey-susanh: Pretty sure it's all or nothing, but I've not tried. IntelliJ offers an option to do the work for us, but not sure how well this works. Still, it's very tempting.

Screenshot 2024-11-04 at 10 03 45 AM

PR #227, opened 2025-01-07
Update module template and dependency versions

labkey-adam:
#### Rationale
A module's first SQL script should look like an incremental script, not a bootstrap script. This ensures we don't remove any Java upgrade code prematurely. Also, we've used three-digit version numbers for five years. And let's stop creating SQL Server scripts.

Update dependency versions.


labkey-susanh (approved):

src/main/groovy/org/labkey/gradle/plugin/XmlBeans.groovy
```diff
@@ -55,6 +55,7 @@ class XmlBeans implements Plugin<Project>
                     xmlbeans "org.apache.xmlbeans:xmlbeans:${project.xmlbeansVersion}"
                 }
 
+        project.logger.info("XMLBeans version: ${project.xmlbeansVersion}")
```
labkey-adam: @labkey-susanh this seemed like a good idea, but I don't see the logging in the gradle build output. Am I looking in the wrong place?

labkey-susanh: If you want it to always show up, you should use the `quiet` log level. Otherwise, use the `-i` flag when running your gradle command. Also, for compatibility with configuration cache, please use `task.logger` instead of `project.logger`.

labkey-adam: I looked for this, but didn't see any other examples of plugins referencing `task.logger`... how do I get a task from this context?

labkey-susanh: Ah, you're right. You aren't in the task execution at this point. So never mind about that.

PR #229, opened 2025-01-09
Stop accepting deprecated properties

labkey-adam:
#### Rationale
The `module.properties` properties `ModuleDependencies`, `ConsolidateScripts`, and `Version` have been deprecated for years. It's time to stop accepting and accommodating them. For now, throw if they're detected to make their use obvious.

We want `ManageVersion` to default to `false` if not specified. This allows us to detect unsupported upgrade scenarios in more cases. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=47369

Remove some unused SVN references.

### DIFF
--- a/src/main/groovy/org/labkey/gradle/plugin/MultiGit.groovy
+++ b/src/main/groovy/org/labkey/gradle/plugin/MultiGit.groovy
@@ -183,22 +183,13 @@ class MultiGit implements Plugin<Project>
         private Boolean isPrivate = false
         private Boolean isExternal = false
         private Boolean isArchived = false
-        private Boolean isSvn = false
         private Type type = Type.other
         private Project project
         private File enlistmentDir
         private Project rootProject
         private Properties moduleProperties
-        private String dependencies
         private String supportedDatabases = "mssql,pgsql"
         private List<PullRequest> pullRequests
-
-        Repository(Project rootProject, String name, Boolean isSvn)
-        {
-            this.name = name
-            this.isSvn = isSvn
-            setProject(rootProject)
-        }
 
         Repository(Project rootProject, String name, String url, Boolean isPrivate, Boolean isArchived, List<String> topics)
         {
@@ -233,7 +224,6 @@ class MultiGit implements Plugin<Project>
             }
             setProject(rootProject)
         }
-
 
         String getName()
         {
@@ -313,26 +303,6 @@ class MultiGit implements Plugin<Project>
         void setIsArchived(Boolean isArchived)
         {
             this.isArchived = isArchived
-        }
-
-        Boolean getIsSvn()
-        {
-            return isSvn
-        }
-
-        void setIsSvn(Boolean isSvn)
-        {
-            this.isSvn = isSvn
-        }
-
-        String getDependencies()
-        {
-            return dependencies
-        }
-
-        void setDependencies(String dependencies)
-        {
-            this.dependencies = dependencies
         }
 
         String getSupportedDatabases()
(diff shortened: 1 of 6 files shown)


src/main/groovy/org/labkey/gradle/plugin/MultiGit.groovy
```diff
@@ -183,23 +183,14 @@ class MultiGit implements Plugin<Project>
         private Boolean isPrivate = false
         private Boolean isExternal = false
         private Boolean isArchived = false
-        private Boolean isSvn = false
```
labkey-susanh, line 186: But what if we decide to go back to SVN? Git is good, but ... sigh ... nostalgia.

src/main/groovy/org/labkey/gradle/plugin/extension/ModuleExtension.groovy
```diff
@@ -81,25 +79,30 @@ class ModuleExtension
...
-                            " Declare the dependency in the module's build.gradle file instead using the 'modules' configuration." +
-                            " See https://www.labkey.org/Documentation/wiki-page.view?name=gradleDepend for more information."
-
-                if (this.modProperties.get("ConsolidateScripts"))
-                    deprecationMsgs += "The 'ConsolidateScripts' property is no longer supported."
-                if (this.modProperties.get("Version"))
-                    deprecationMsgs += "The 'Version' property is no longer supported."
+//          Follow this pattern to deprecate a property in module.properties
+//                if (this.modProperties.get("OldProperty"))
+//                    deprecationMsgs += "The OldProperty property is no longer supported."
```
labkey-susanh, line 89: Thanks for leaving this behind here.

src/main/groovy/org/labkey/gradle/plugin/extension/ModuleExtension.groovy
```diff
@@ -81,25 +79,30 @@ class ModuleExtension
...
                     project.logger.quiet("${propertiesFile.absolutePath}: Deprecated or unsupported properties detected.\n\t"
-                            + deprecationMsgs.join("\n\t")
-                            + "\nRefer to https://www.labkey.org/Documentation/wiki-page.view?name=includeModulePropertiesFile for the current set of supported properties.")
+                        + deprecationMsgs.join("\n\t")
+                        + "\nRefer to https://www.labkey.org/Documentation/wiki-page.view?name=includeModulePropertiesFile for the current set of supported properties.")
             }
-            if (this.modProperties.get(MODULE_DEPENDENCIES_PROPERTY))
-                this.modProperties.remove(MODULE_DEPENDENCIES_PROPERTY)
+
+            List<String> unsupportedMsgs = []
```
labkey-susanh, line 96: Can we maybe leave a note here about when we can remove these checks altogether? Seems like probably the next time we change the version of LabKey we're compatible with would be sufficient.

labkey-susanh (approved):
Changes seem good here. Please be sure to run a build on TeamCity with these changes at least once before merge.

src/main/groovy/org/labkey/gradle/plugin/MultiGit.groovy
labkey-adam: I'd advocate for CVS

src/main/groovy/org/labkey/gradle/plugin/extension/ModuleExtension.groovy
labkey-adam: Good idea. It's difficult to decide on timeframe because we don't know who is affected (if anyone), how often they build their modules, and what release they're building against (they'll see the warnings & exceptions only if they're building on a branch with the new plugins version). My compromise proposal: we deprecate for one year and throw for one year after that. LMK if you disagree.

And thanks for reminding me: since I'm switching the default value of `ManageVersion`, the next plugins version will require 25.2+.

labkey-susanh: Your compromise proposal works for me.

labkey-adam:
Build and DRTs are successful: https://github.com/LabKey/server/pull/964
Installer build is successful: https://teamcity.labkey.org/buildConfiguration/LabKey_Trunk_Premium_InstallersCustom/3338386

PR #230, opened 2025-01-24
Don't log XMLBeans version so loudly

labkey-tchad:
#### Rationale
Gradle's "quiet" [log level](https://docs.gradle.org/current/userguide/logging.html) is actually the second loudest, between "error" and "warn". It is also the default log level. We probably don't need to log the XMLBeans version a dozen times for normal builds.
Do we need it at all? Maybe this was accidentally left in.

#### Related Pull Requests
- #227 

#### Changes
- Log XMLBeans version at a lower level


labkey-tchad:
Looks like this is known and intentional: https://github.com/LabKey/server/pull/966

PR #233, opened 2025-08-01
Update `PurgeArtifacts` task for configuration cache and more complete clean up

labkey-susanh:
#### Rationale
The `PurgeArtifacts` task was walking through the set of gradle projects to find what needed to be purged, which is fine until we archive repositories so they are no longer needed in a typical enlistment. This implementation, because it used `project.allprojects`, is also not compatible with Gradle's configuration cache. This PR changes the implementation to read the list of projects from a file and also will read a list of versions from a file to make it easier to purge several versions. 

#### Related Pull Requests
- N/A. This is a task that is not run on TeamCity, so testing there is not possible/necessary. I've used this to good end locally for several recent purges.

#### Changes
- Update `PurgeArtifacts` with more configuration cache-friendly properties
- Update logic to read multiple options from files instead of single options


labkey-tchad (approved):

gradle.properties
```diff
@@ -3,7 +3,7 @@ artifactory_contextUrl=https://labkey.jfrog.io/artifactory
 artifactoryPluginVersion=5.2.5
 
 commonsIoVersion=2.18.0
-commonsLang3Version=3.17.0
+commonsLang3Version=3.18.0
```
labkey-adam, line 6: Consider updating all the other dependencies... I believe Commons IO, Commons Text, HttpClient, HttpCore, Jackson, and Json all have newer versions.

labkey-adam (approved):

gradle.properties
labkey-susanh: Done. Everything except `artifactoryPluginVersion`, where the latest version requires Gradle 9.

PR #234, opened 2025-09-17
Update to Gradle 9 and make more progress toward configuration cache compatibility

labkey-susanh:
#### Rationale
[Gradle 9.0.0](https://docs.gradle.org/9.0.0/release-notes.html) has been released (and 9.1.0 is coming soon). This release has removed some deprecated features and moves even closer to making the use of the configuration cache the default behavior, so getting out build closer to compatibility will help make the transition easier whenever that happens (Probably Gradle 10, but I'm unsure on the timing of that).

#### Related Pull Requests
- https://github.com/LabKey/server/pull/1182
- https://github.com/LabKey/premiumModules/pull/357
- https://github.com/LabKey/DiscvrLabKeyModules/pull/391
- https://github.com/LabKey/wnprc-modules/pull/878
- https://github.com/LabKey/testAutomation/pull/2695

#### Changes
- Update DoThenSetup and its relatives to be compatible with configuration cache
- Update DeployApp for better configuration cache compatibility
- Replace usages of project.javaexc with JavaExc task types or injected ExecOperations
- Update CompressClientLibs, StopLabKey to work better with configuration cache
- Remove `listNodeModules` task defined by `NpmRun` plugin
- Remove `cleanOut` task defined by `FileModule` plugin
- Remove `cleanAndDeploy` task from `ServerDeploy` plugin
- Update `CopyAndInstallRPackage` Task class to use injected FileSystem instead of `project.copy`
- Update `GzipAction` task to use task logger and ant builder objects
- Update `StopLabKey` and `StartLabKey` with appropriate dependencies to mirror the deprecated `StopTomcat` and `StartTomcat`
- Remove deprecated `getServerDeployDirectory` from `ServerDeployExtension`
- Remove `createNlpConfig` task from `TeamCity` plugin since that module is no longer in use
- Remove `setup` task from `ServerDeploy` plugin.
- Update `DeployDistribution` and `StageDistribution` for configuration cache compatibility
- Some updates for `ModuleDistribution` for better configuration cache compatibility
- Update `PropertiesUtils` to remove code for supporting `labkey.xml`
- Relocate the `.restartTriggerFile` to prevent creation of empty `build/deploy/modules` directory for embedded distribution deployment
- Fix `bootstrap` task so it will update `application.properties`


src/main/groovy/org/labkey/gradle/plugin/extension/DistributionExtension.groovy
```diff
@@ -28,8 +28,7 @@ class DistributionExtension
     public static final String TAR_ARCHIVE_EXTENSION = "tar.gz"
 
     String dir = "${project.rootProject.projectDir}/dist"
-    String artifactId
-    String description
+    String description // TODO this is set in various distribution build.gradle files but doesn't seem to be used. Remove?
```
labkey-tchad: Distributions have been updated to not use this. Probably safe to remove it, assuming we are going to bump the LabKey version compatibility.

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
```diff
@@ -209,6 +196,12 @@ class FileModule implements Plugin<Project>
                 }
             }
 
+        project.tasks.named('deployModule')
+            {
+                notCompatibleWithConfigurationCache("Needs its own class to do the two copies (one to staging and one to deploy or possibly two Copy tasks chained together.")
+            }
```
labkey-tchad: Why is this configuration separate from the registration of the `deployModule` task just above this?

src/main/groovy/org/labkey/gradle/plugin/ServerDeploy.groovy
```diff
@@ -190,48 +156,64 @@ class ServerDeploy implements Plugin<Project>
...
+                    task.dependsOn(project.tasks.npmSetup)
+                    task.notCompatibleWithConfigurationCache("Needs its own class to declare proper input and output properties")
+            }
+            project.tasks.symlinkNode.notCompatibleWithConfigurationCache("References project properties. Need to add task class with input properties")
+            project.tasks.named('deployApp').configure {dependsOn(project.tasks.symlinkNode)}
         }
 
+
         if (BuildUtils.embeddedProjectExists(project)) {
             def embeddedProject = project.project(BuildUtils.getEmbeddedProjectPath())
```
labkey-tchad, line 199: Is `BuildUtils.embeddedProjectExists` even relevant anymore? (probably out of scope but something to clean up later)

src/main/groovy/org/labkey/gradle/task/TeamCityPropertiesTask.groovy
```diff
@@ -0,0 +1,147 @@
...
+
+    static String getLabKeyContextPath(Project project)
+    {
+        return getTeamCityProperty(project, "labkey.contextpath", null)
+    }
+
+    static String getLabKeyServerPort(Project project)
+    {
+        return getTeamCityProperty(project, 'tomcat.port', null)
+    }
```
labkey-tchad: Putting these static TeamCity property getters in the task class seems odd. It also causes some method name overlap; there is a static `getLabKeyServerPort` that gets the property but there is also the implicit non-static `getLabKeyServerPort` for the task input.
If `TeamCityExtension` isn't the right place for these, maybe they could live in a new `TeamCityUtils` (similar to `BuildUtils`).

labkey-tchad (approved):

src/main/groovy/org/labkey/gradle/plugin/FileModule.groovy
labkey-susanh: No good reason, just the mood I was in at the time. Moved

src/main/groovy/org/labkey/gradle/plugin/ServerDeploy.groovy
labkey-susanh: It is still technically possible that you won't have the embedded project included in your `settings.gradle` file. I'm not sure why you would, but it's possible.

src/main/groovy/org/labkey/gradle/task/TeamCityPropertiesTask.groovy
labkey-susanh: The placement of these methods doesn't seem odd to me, but good point about the name overlap. I don't remember the evolution of this change, but I am generally moving away from heavy usage of Extension classes where possible because we don't seem to be using them in the way they are intended in a lot of places. No real harm for static classes, though, so I've moved these methods back (for now, at least).

PR #235, opened 2025-10-15
Update StartLabKey and StopLabKey tasks

labkey-tchad:
#### Rationale
[Issue 53792: Capture heap dumps when a test causes an OutOfMemoryError on TeamCity](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53792)
[Issue 53244: Update 'stopTomcat' Gradle task to work without Spring's shutdown endpoint](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=53244)

#### Related Pull Requests
N/A

#### Changes
- Capture heap dumps when a test causes an OutOfMemoryError on TeamCity
- Update 'stopTomcat' Gradle task to work without Spring's shutdown endpoint


src/main/groovy/org/labkey/gradle/task/StopLabKey.groovy
```diff
@@ -16,40 +16,50 @@
...
-        def endpoint = "${useSsl.get() ? "https" : "http"}://localhost:$port/actuator/shutdown"
-        def command = "curl -X POST $endpoint"
-        this.logger.info("Sending command to $endpoint")
-        def proc = command.execute()
-        proc.waitFor()
-        if (proc.exitValue() != 0)
-            this.logger.warn("Shutdown command exited with non-zero status ${proc.exitValue()}.")
-        else
-            this.logger.quiet("Shutdown successful")
+        File pidFile = ServerDeployExtension.getEmbeddedDir(project).file("labkey.pid").asFile
```
labkey-susanh: Per current Gradle conventions, this should be designated as an `@InputFile`. Even though this task isn't going to benefit from the configuration cache, it seems better to use this convention and not have to worry about designating this as not compatible with the configuration cache. (I believe in Gradle 10, or thereabouts, tasks may fail if they aren't compatible or don't explicitly declare themselves incompatible.)

labkey-susanh (changes requested):

src/main/groovy/org/labkey/gradle/task/StopLabKey.groovy
labkey-tchad: I tried that initially but it was throwing errors if the pid file didn't exist. I think I've figured it out though (Optional doesn't mean what I thought it meant)

labkey-susanh (approved):

PR #238, opened 2025-12-08
Remove extra logging, add additional task for removing npm versions and remove use of versioning plugin

labkey-susanh:
#### Rationale
- We no longer need to assure ourselves that we are using the intended version of XMLBeans
- A more general task for purging NPM versions from Artifactory helps to keep things tidy
- We can get the Git data we need for the `module.xml` file without the (dying) versioning plugin

#### Related Pull Requests
- https://github.com/LabKey/server/pull/1229

#### Changes
- Remove logging of XMLBeans version
- Add `PurgeNpmVersions` task
- Remove use of versioning plugin in favor of some simple git commands


labkey-tchad (approved):
If this doesn't work on Windows, maybe we could use the JGit library used by our VCS license tool? https://github.com/LabKey/tools/blob/develop/vcslicense/src/org/labkey/vcslicense/VCSLicense.java

labkey-susanh:
> If this doesn't work on Windows, maybe we could use the JGit library used by our VCS license tool? https://github.com/LabKey/tools/blob/develop/vcslicense/src/org/labkey/vcslicense/VCSLicense.java

Good thought to keep in mind. Fortunately, indications are that it does work on Windows as currently written.

README.md
```diff
@@ -13,12 +13,15 @@ on how to do that, including how to develop and test locally and the versioning
 
 ## Release Notes
 
-### TBD
+### 7.X.X
 *Released*: TBD
 (Earliest compatible LabKey version: 25.10)
```
labkey-adam, line 18: 25.12?

labkey-adam (approved):

README.md
labkey-susanh: I think this should still be compatible with 25.10 since the presence of the versioning plugin version shouldn't cause problems; we simply won't be using it.

PR #240, opened 2026-01-07
GitHub Issue 774: Fix distribution file name for snapshot versions

labkey-susanh:
#### Rationale
Issue [774](https://github.com/LabKey/internal-issues/issues/771): When we removed the usage of the versioning plugin, we missed a dependency on that plugin that was used for determining the names of our distribution files.

The `RestoreFromTrash` task needed converting to be compatible with the configuration cache and to allow for a more robust restoration process.

The `PurgeArtifacts` task was not checking for empty versions, and if purge requests are sent with an empty version, that causes all versions to be deleted. Not something we want.

#### Related Pull Requests
- https://github.com/LabKey/server/pull/1256

#### Changes
- Update `BuildUtils.getDistributionVersion` to get VCS properties from git commands instead of removed versioning plugin
- Add some trimming and empty string checks for `PurgeArtficats` task
- Modify `RestoreFromTrash` task to read modules and versions from files

### DIFF
--- a/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
+++ b/src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
@@ -401,10 +401,11 @@ class BuildUtils
     {
         String version = project.labkeyVersion
 
-        if (project.hasProperty("versioning"))
+        if (project.hasProperty("includeVcs"))
         {
-            String branch = project.versioning.info.branchId
-            if (!["trunk", "develop", "master", "main", "none", ""].contains(branch) &&
+            var vcsProps = getStandardVCSProperties(project)
+            String branch = vcsProps.get(VCS_BRANCH_PROP_NAME)
+            if (branch != null && !["trunk", "develop", "master", "main", "none", ""].contains(branch) &&
                     !branch.toLowerCase().matches("release.*-snapshot"))
             {
                 Matcher matcher = Pattern.compile(".*fb_(.+)").matcher(branch)
@@ -601,10 +602,7 @@ class BuildUtils
 
     static void addModuleDistributionDependency(Project distributionProject, String depProjectPath, String config, boolean addTransitive)
     {
-        if (distributionProject.configurations.named(config) == null)
-            distributionProject.configurations {
-                config
-            }
+        distributionProject.configurations.maybeCreate(config)
         distributionProject.logger.info("${distributionProject.path}: adding ${depProjectPath} as dependency for config ${config}")
         addLabKeyDependency(project: distributionProject, config: config, depProjectPath: depProjectPath, depProjectConfig: "published", depExtension: "module", depVersion: distributionProject.labkeyVersion)
         if (addTransitive) {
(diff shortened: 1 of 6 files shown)


src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -594,7 +601,7 @@ class BuildUtils
 
     static void addModuleDistributionDependency(Project distributionProject, String depProjectPath, String config, boolean addTransitive)
     {
-        if (distributionProject.configurations.findByName(config) == null)
+        if (distributionProject.configurations.named(config) == null)
             distributionProject.configurations {
                 config
             }
```
labkey-tchad: I think this could use [`maybeCreate`](https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ConfigurationContainer.html#org.gradle.api.artifacts.ConfigurationContainer:maybeCreate(java.lang.String)).
```
distributionProject.configurations.maybeCreate(config)
```

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
```diff
@@ -613,16 +620,16 @@ class BuildUtils
             return
 
         distributionProject.evaluationDependsOn(depProject.getPath())
-        if (depProject.configurations.findByName("modules") != null) {
+        if (depProject.configurations.named("modules") != null) {
```
labkey-tchad: I don't think this will work the same way. [`named`](https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ConfigurationContainer.html#org.gradle.api.artifacts.ConfigurationContainer:named(java.lang.String)) throws an exception instead of returning `null`.
There's no advantage to using lazy resolution here since we're immediately digging in to the configuration's contents, right?

labkey-tchad (changes requested):
`BuildUtils.getVersionNumber` also references the versioning plugin. Could you update that as well? If I remember correctly, it's used to include feature branch names in `module.xml` files for modules built from a feature branch.

labkey-tchad:
After some investigation, it turns out `BuildUtils.getVersionNumber` was included in `.module` file names before removing the versioning plugin.

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-susanh: Thanks. Done.

src/main/groovy/org/labkey/gradle/util/BuildUtils.groovy
labkey-susanh: Ah, you are so right. I forgot about that. (You would think I might remember since it's really annoying that this throws an exception).

labkey-susanh:
> After some investigation, it turns out `BuildUtils.getVersionNumber` was included in `.module` file names before removing the versioning plugin.

I've restored the functionality for now. I think it's still at least theoretically desirable, even if we are seldom using `-PincludeVcs` with feature branches.

labkey-tchad (approved):

PR #242, opened 2026-04-02
Issue 464: Prevent multiple versions of jars getting into distributions & Issue 1015 (special treatment for some classifier versions)

labkey-susanh:
#### Rationale
Issue [464](https://github.com/LabKey/internal-issues/issues/464)
Issue [1015](https://github.com/LabKey/internal-issues/issues/1015)

#### Related Pull Requests
- https://github.com/LabKey/server/pull/1324

#### Changes
- Update gradle version
- Update dependency versions
- Add attributes for the `apiJarFile` and `published` configurations to be able to distinguish the two
- Update `Distribution.substituteModuleDependencies` to use attributes for selecting the `.module` files instead of the `.jar` files when doing substitution
- Update `CheckForVersionConflicts` with some hacky logic to treat some classified versions as versions that conflict with each other.
- Remove `VcsTag` from module.xml file since it is not being populated with a useful value


src/main/groovy/org/labkey/gradle/task/CheckForVersionConflicts.groovy
```diff
@@ -34,6 +34,40 @@ class CheckForVersionConflicts  extends DefaultTask
     @Input
     Set<String> MULTIPLE_VERSIONS_ALLOWED = Set.of("jackson-core", "jackson-databind")
 
+    // GH Issue 1015: We are using milestone versions of spring-ai jars, which use classifiers like -M2 to distinguish the differnet versions.
+    // We want to have the later milestones replace the earlier ones, so we want to exclude the milestone classifier from the name when
+    // comparing for conflicts. This is rather sketchy and I hope it goes away soon. The list has to include all transitive dependencies as
+    // we as the direct dependencies, and since the artifacts are not released for production use they are not entirely stable, so it's entirely
```
labkey-adam:
```suggestion
    // well as the direct dependencies, and since the artifacts are not released for production use they are not entirely stable, so it's entirely
```

labkey-adam (approved):

src/main/groovy/org/labkey/gradle/task/CheckForVersionConflicts.groovy
```diff
@@ -34,6 +34,40 @@ class CheckForVersionConflicts  extends DefaultTask
...
+    Set<String> USE_CLASSIFIER_IN_VERSION = Set.of(
+            "spring-ai-bom",
+            "spring-ai-starter-model-google-genai",
+            "spring-ai-anthropic",
+            "spring-ai-openai",
+            "spring-ai-client-chat",
+            "spring-ai-advisors-vector-store",
+            "spring-ai-starter-model-google-genai-embedding",
+            "spring-ai-autoconfigure-model-chat-client",
+            "spring-ai-autoconfigure-mcp-sever-common",
```
labkey-jeckels:
```suggestion
            "spring-ai-autoconfigure-mcp-server-common",
```

src/main/groovy/org/labkey/gradle/task/CheckForVersionConflicts.groovy
```diff
@@ -98,7 +133,7 @@ class CheckForVersionConflicts  extends DefaultTask
                     }
                     else
                     {
-                        haveMultiples = true
+                        haveMultiples = !USE_CLASSIFIER_IN_VERSION.contains(nameWithoutClassifier)
```
labkey-jeckels: This will currently remember the value from the last time through the loop. Maybe it'll always be the same, but seems safer to `or` with previous values.

```suggestion
                        haveMultiples = haveMultiples || !USE_CLASSIFIER_IN_VERSION.contains(nameWithoutClassifier)
```

src/main/groovy/org/labkey/gradle/task/CheckForVersionConflicts.groovy
```diff
@@ -34,6 +34,40 @@ class CheckForVersionConflicts  extends DefaultTask
     @Input
     Set<String> MULTIPLE_VERSIONS_ALLOWED = Set.of("jackson-core", "jackson-databind")
 
+    // GH Issue 1015: We are using milestone versions of spring-ai jars, which use classifiers like -M2 to distinguish the differnet versions.
```
labkey-jeckels:
```suggestion
    // GH Issue 1015: We are using milestone versions of spring-ai jars, which use classifiers like -M2 to distinguish the different versions.
```

src/main/groovy/org/labkey/gradle/task/CheckForVersionConflicts.groovy
```diff
@@ -34,6 +34,40 @@ class CheckForVersionConflicts  extends DefaultTask
...
+    // We want to have the later milestones replace the earlier ones, so we want to exclude the milestone classifier from the name when
+    // comparing for conflicts. This is rather sketchy and I hope it goes away soon. The list has to include all transitive dependencies as
+    // well as the direct dependencies, and since the artifacts are not released for production use they are not entirely stable, so it's entirely
+    // possible that, say, a jar with milestone M2 will not have the same name with M3 or M4 and won't get cleaned up during the conflict checking.
+    @Input
+    Set<String> USE_CLASSIFIER_IN_VERSION = Set.of(
```
labkey-jeckels: Could we match based on a `spring-ai-` prefix instead of the full list?

labkey-susanh: Could. It's all a bit of a hack, so ..

labkey-susanh: @labkey-adam Do you have opinions on whether matching with a prefix would be more or less problematic here?

labkey-adam: All 25 `spring-ai-*.jar` files in my deployment have the `M4` classifier, so I suspect the prefix approach would be reliable and more straightforward

src/main/groovy/org/labkey/gradle/task/CheckForVersionConflicts.groovy
labkey-susanh: Thanks. Done.

src/main/groovy/org/labkey/gradle/task/CheckForVersionConflicts.groovy
labkey-susanh: Done.

labkey-susanh: Done.

PR #245, opened 2026-05-18
Fix caching issue for `ClientLibsCompress` and updates for Issue 1143: Remove GWT-related code 

labkey-susanh:
#### Rationale
Issue [1143](https://github.com/LabKey/internal-issues/issues/1143): We have removed all usages of GWT in the code so no longer need the plugin or extension.

`ClientLibsCompress` had improper logic related to its declaration of output files, leading to some intermittent failures on TeamCity when the task got annotated as a `CacheableTask`

[Gradle 9.5.1](https://docs.gradle.org/9.5.1/release-notes.html) is the latest

#### Related Pull Requests
- https://github.com/LabKey/server/pull/1378

#### Changes
- Update to Gradle 9.5.1
- Remove GWT plugin and extension
- Declare `ClientLibsCompress` task as not cacheable (for now)


labkey-jeckels (approved):

labkey-tchad (changes requested):
I spent a while on Friday trying to repro and fix the caching problem with Claude.
In addition to this change to `ClientLibsCompress`, it suggested excluding `XmlImporter`s from the list of output files when they have `doCompile==false`. This seems correct since `compressSingleFile` also skips such importers.
It also suggested declaring `isDevMode` as an input for that task so that dev and production builds don't share cache keys.

labkey-susanh:
> I spent a while on Friday trying to repro and fix the caching problem with Claude. In addition to this change to `ClientLibsCompress`, it suggested excluding `XmlImporter`s from the list of output files when they have `doCompile==false`. This seems correct since `compressSingleFile` also skips such importers. It also suggested declaring `isDevMode` as an input for that task so that dev and production builds don't share cache keys.

Done. Our conversations with Claude were clearly a bit different. 

I've also removed the "combined" file as an output since I can't find reference to it anywhere or remember where it was needed. If you or anyone you know recognizes it as necessary we can happily restore it.

src/main/groovy/org/labkey/gradle/task/ClientLibsCompress.groovy
```diff
@@ -44,8 +47,9 @@ import java.nio.charset.StandardCharsets
 import java.util.stream.Collectors
 
 /**
- * Class for compressing javascript and css files using the yuicompressor classes.
+ * Class for compressing javascript and css files using terser
  */
+//@DisableCachingByDefault(because="Troubleshooting")
```
labkey-tchad:
```suggestion
```

labkey-tchad (approved):

